PR #1, opened 2015-09-16
Fix path to server files

justineaster:
The old path is from last year's server location.


prattmic:
Whoops, that page is woefully out of date. Even with the appropriate path, the tests won't work properly, because the server runs in the virtualenv.

https://github.com/auvsi-suas/interop/blob/master/setup/README.md has up-to-date testing instructions. I can fix this page later today or tomorrow.

prattmic:
This is fixed by #3

PR #2, opened 2015-09-16
Fix path to server and add activation command

justineaster:
Runserver will not work unless venv has been activated.


prattmic:
LGTM

PR #6, opened 2015-09-18
Enable and fix load tests

### LINKED ISSUE #5: Testing script does not print operation rate
In docs/setup/testing.rst it states:
"Each interaction must operate at 10 Hz while doing 3 operations simultaneously, so 30 Hz each in isolation. The rates achieved will be printed to the screen"

As far as I could tell the testing script does not actually print the operation rate.

prattmic:
Enable the load tests by default and fix the telemetry load test, which
was missing an import.

Fixes #5


pmtischler:
LGTM

PR #12, opened 2015-11-05
Documentation update

prattmic:
Several parts of the documentation are out-of-date and incorrect.
Correct them.

There is also some missing information, such as the default username and
password to document,


pmtischler:
LGTM

PR #15, opened 2015-11-07
Add Travis CI config

prattmic:
The config uses the exact same Puppet setup as a normal installation,
and runs the typical tests. It takes about 5 minutes to run.

In October, Travis finally added 14.04 as a base distro. My previous attempts to get our automated setup working on 12.04 were painful to say the least.


pmtischler:
LGTM

PR #16, opened 2015-11-07
Install PyYAML for server

### LINKED ISSUE #7: Username and Pasword
While running the file i'm having this particular issue i.e. username and password invalid. I'm using the default values as username and password i.e. username,password and also tried testadmin and testpass.
Altthough i am able to run the judges server but can't run the client server. both the clients failed.
please help ASAP

prattmic:
PyYAML is required for ``loaddata`` to be able to load our test fixture
into the database.

We were missing this dependency before, and simply relied on the system
having it installed by default.

Fixes #7

### DIFF
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -7,6 +7,7 @@ matplotlib
 numpy
 pillow
 psycopg2
+PyYAML # for loaddata
 scipy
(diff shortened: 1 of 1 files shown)


pmtischler:
LGTM

PR #17, opened 2015-11-07
Create a setup log file

prattmic:
In addition to displaying setup output on the screen, redirect it to a
log file. This log will be helpful if teams need to upload setup output
for support.

### DIFF
--- a/setup/setup.sh
+++ b/setup/setup.sh
@@ -34,6 +34,9 @@ function ensure_puppet_module() {
     sudo puppet module upgrade ${module}
 }
 
+# Save all output to a log file.
+exec &> >(tee ${SETUP}/setup-$(date +%F-%H-%M-%S).log)
+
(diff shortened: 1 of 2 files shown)


setup/setup.sh
```diff
@@ -34,6 +34,9 @@ function ensure_puppet_module() {
     sudo puppet module upgrade ${module}
 }
 
+# Save all output to a log file.
+exec &> >(tee ${SETUP}/setup-$(date +%F-%H-%M-%S).log)
```
pmtischler: Do we want to have it here or right after the `SETUP` variable is defined?

pmtischler:
LGTM

setup/setup.sh
prattmic: There is no work done earlier, just function definitions.

Issue #13, opened 2015-11-06
Duplicate GpsPositions added


legonigel:
After running the test server for some time it was severely slowed down. I believe this is related to the rediculous number of GpsPosition items. Currently my copy of the server has 201064 gpsposition items. More than 10% of these items are duplicate of a gps position already in the database.  I stopped checking for duplicates at 10% because I confirmed my suspicion of a large number of duplicate gps positions. 

I have not looked at the code enough to verify this, but I believe most of these GpsPosition elements come from the posted telemetry items.

pmtischler:
I'm happy to help debug the issue with you, but let's not jump to conclusions just yet :)

To my knowledge, we only save a `GpsPosition` in:
* [UAS telemetry upload](https://github.com/auvsi-suas/interop/blob/9b98df13502caa746e7ba7f6ce84d3c8afd5f400/server/auvsi_suas/views/telemetry.py#L75)
* [Target submission](https://github.com/auvsi-suas/interop/blob/9b98df13502caa746e7ba7f6ce84d3c8afd5f400/server/auvsi_suas/views/targets.py#L151)

Neither code paths contain a loop or multiple save calls. Our code does not dedupe `GpsPositions` on write, it dedupes the positions on read (e.g. as part of post-mission evaluation), so it is entirely possible you are uploading duplicate data.

201064 positions is not a lot. At 10Hz that is ~5.6 hours of uploads, we sustained longer at competition without any issue. The `GpsPosition` model has 3 fields which are at most 8 bytes each (id, latitude, longitude), so that is only ~4.6MB, a very small size. The Postgres database can handle much larger. We [set our database-side caching to 1GB](https://github.com/auvsi-suas/interop/blob/9b98df13502caa746e7ba7f6ce84d3c8afd5f400/setup/puppet_files/auvsi_suas/manifests/postgresql_setup.pp#L28), so serving should be cached in memory completely. With database journaling, our [disabling of synchronous commits via fsync](https://github.com/auvsi-suas/interop/blob/9b98df13502caa746e7ba7f6ce84d3c8afd5f400/setup/puppet_files/auvsi_suas/manifests/postgresql_setup.pp#L50), and [512MB window for checkpointing](https://github.com/auvsi-suas/interop/blob/9b98df13502caa746e7ba7f6ce84d3c8afd5f400/setup/puppet_files/auvsi_suas/manifests/postgresql_setup.pp#L38), we shouldn't have any issues dealing with 4.6MB of writes.

[...]

PR #18, opened 2015-11-07
Display raw ID fields for models with many instances

prattmic:
Certain models, such as GpsPosition and AerialPosition may have
thousands or hundreds of thousands of entries once a significant amount
of telemetry has been to the server.

By default, the admin interface will enumerate all options for these
ForeignKeys. This is infeasible when there are too many instances, so
just display a raw ID field instead. There will be a search button next
to the box, where a specific instance can be searched for or created.

References #13


pmtischler:
LGTM

PR #19, opened 2015-11-07
Verify docs and formatting on Travis CI

prattmic:
This PR consists of several commits.

First I extended ``tools/format.sh`` with more options so that Travis could check all files.

Then I updated and froze ``yapf``, re-running the latest version on all of the files. Since it is a bit finicky from version to version, I have frozen it at the current latest version.

Finally, I added a script to ensure the docs build and enabled both in the Travis config.

### DIFF
--- a/client/interop/types.py
+++ b/client/interop/types.py
@@ -94,8 +94,8 @@ class ServerInfo(Serializable):
     """
 
     def __init__(self, message, message_timestamp, server_time):
-        super(ServerInfo, self).__init__(
-            ["message", "message_timestamp", "server_time"])
+        super(ServerInfo, self).__init__(["message", "message_timestamp",
+                                          "server_time"])  # yapf: disable
 
         self.message = message
         self.message_timestamp = dateutil.parser.parse(message_timestamp)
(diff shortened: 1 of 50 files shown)


client/interop/types.py
```diff
@@ -94,8 +94,8 @@ class ServerInfo(Serializable):
     """
 
     def __init__(self, message, message_timestamp, server_time):
-        super(ServerInfo, self).__init__(
-            ["message", "message_timestamp", "server_time"])
+        super(ServerInfo,
+              self).__init__(["message", "message_timestamp", "server_time"])
```
pmtischler: Is this what we want it to look like?

server/auvsi_suas/views/targets_test.py
```diff
@@ -625,8 +633,8 @@ def test_get_no_image(self):
 
     def test_delete_no_image(self):
         """404 when DELETE image before upload."""
-        response = self.client.delete(
-            targets_id_image_url(args=[self.target_id]))
+        response = self.client.delete(targets_id_image_url(args=[self.target_id
+                                                                 ]))
```
pmtischler: Same here.

pmtischler:
LGTM

client/interop/types.py
prattmic: No, but I couldn't convince yapf to do anything else. I fixed it and disabled formatting.

server/auvsi_suas/views/targets_test.py
prattmic: Same

PR #20, opened 2015-11-07
Verify that target requests are valid JSON

prattmic:
Currently, bad JSON results in an exception and 500 error to the client.

### DIFF
--- a/server/auvsi_suas/views/targets_test.py
+++ b/server/auvsi_suas/views/targets_test.py
@@ -200,6 +200,13 @@ class TestPostTarget(TestCase):
                                     content_type='application/json')
         self.assertEqual(400, response.status_code)
 
+    def test_invalid_json(self):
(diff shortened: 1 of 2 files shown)


server/auvsi_suas/views/targets_test.py
```diff
@@ -200,6 +200,13 @@ def test_missing_type(self):
                                     content_type='application/json')
         self.assertEqual(400, response.status_code)
 
+    def test_invalid_json(self):
```
pmtischler: Nit: `test_post_invalid_json`

pmtischler:
LGTM

server/auvsi_suas/views/targets_test.py
prattmic: This is the TestPostTarget class, all of the cases are POST.

PR #27, opened 2016-01-23
Avoid the latest virtualenv wheel

### LINKED ISSUE #26: Python Installation
Attempting to install Python onto our based system as one of the prerequisites. We are unable to pass the initial test. How should we go about installing Python 2 and Python 3 to pass this test?

prattmic:
The virtualenv 14.0.1 package wheel creates a virtualenv-3.5 binary,
regardless of Python version. Since we are running Python 3.4, this
breaks later setup steps. By avoiding the wheel (and using the tar
instead), we avoid the issue.

See the virtualenv issue at pypa/virtualenv#851
(https://github.com/pypa/virtualenv/issues/851).

Fixes #26


pmtischler:
LGTM

Issue #28, opened 2016-01-26
Interpreter lines (#! lines) don't work with Windows+Vagrant


pmtischler:
Change all scripts so they don't use #! but rather execute scripts with bash/python.

PR #39, opened 2016-01-28
Remove interpreter line execution to support Vagrant-Windows.

pmtischler:
Execute scripts with interpreter type.

For #28

### DIFF
--- a/docs/setup/testing.rst
+++ b/docs/setup/testing.rst
@@ -10,13 +10,13 @@ To run the tests execute the following commands:
 
 .. code-block:: bash
 
-    $ ~/interop/test.sh
+    $ bash /interop/test.sh
 
 Or from outside Vagrant VM:
 
 .. code-block:: bash
 
-    $ vagrant ssh -c "/interop/test.sh"
+    $ vagrant ssh -c "bash /interop/test.sh"
 
 The tests have executed successfully if you see these lines::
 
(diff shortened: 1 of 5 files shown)


prattmic:
Don't actually remove the shebangs, they are useful when running manually, though removing them does provide a nice way to test.

Are there no Python scripts that need to be changed?

Note that the tests are failing.

pmtischler:
I've put the shebangs back. I don't believe we call any python files or bash scripts directly, other than in Vagrant setup (or equivalently our doc for machine setup) and in the `test.sh` file.

It appears the tests are failing due to docs not building, which is happening for both PRs and seems to be a new error unrelated to this PR.

prattmic:
`docs/setup/testing.rst` has direct invocations of test.sh.

pmtischler:
Updated.

docs/setup/testing.rst
```diff
@@ -10,7 +10,7 @@ To run the tests execute the following commands:
 
 .. code-block:: bash
 
-    $ ~/interop/test.sh
+    $ bash /interop/test.sh
 
 Or from outside Vagrant VM:
```
prattmic: The `vagrant ssh` invocation below also needs to be updated.

prattmic:
Also `setup/README.md`

pmtischler:
Done & Done.

prattmic:
LGTM

PR #42, opened 2016-01-31
Work around sphinx-doc/sphinx#2280

### LINKED ISSUE #41: Docs build failing, causing Travis tests to fail
Example:
https://travis-ci.org/auvsi-suas/interop/builds/105410327

prattmic:
The docs are currently building with warnings because one HTTP
sourcecode block cannot be highlighted. The root cause is
sphinx-doc/sphinx#2280. Once that is fixed, the block should be fine as
previously written. Until then, explicitly add a Content-Type line to
avoid the bug.

Fixes #41

### DIFF
--- a/docs/spec.rst
+++ b/docs/spec.rst
@@ -778,9 +778,13 @@ Targets
 
    **Example response**:
 
+   .. FIXME(sphinx-doc/sphinx#2280): The Content-Type here is not particularly
+      relevant, but otherwise the sourcecode block will fail to lex.
+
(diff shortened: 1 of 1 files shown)


pmtischler:
LGTM.

prattmic:
The underlying problem is fixed in Pygments. We can revert this workaround once they release a new version.

https://bitbucket.org/birkenfeld/pygments-main/commits/5d46d24abbc4ce1aa4d4e347a5da67f4e040372e

Issue #36, opened 2016-01-27
Log all requests and responses


pmtischler:
From @prattmic:

It would be nice to have one central mechanism that logged all users' requests and the associated response, so we can better keep track of what is going on.

Something along these lines: http://stackoverflow.com/a/12942470/10817

PR #43, opened 2016-01-31
Add logging middleware. Configure logging.

pmtischler:
For #36

### DIFF
--- a/server/server/settings.py
+++ b/server/server/settings.py
@@ -51,6 +51,7 @@ MIDDLEWARE_CLASSES = (
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'auvsi_suas.views.middleware.LoggingMiddleware',
 )  # yapf: disable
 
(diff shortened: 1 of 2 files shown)


server/server/settings.py
```diff
@@ -115,14 +116,25 @@
         },
     },
     'loggers': {
+        'py.warnings': {
```
prattmic: What do all of these categories do?

pmtischler: These are the named loggers. For example, the named logger we created [here](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/views/__init__.py#L3) is configured by the category below.

prattmic: Is there a way to define 'default' settings for loggers not specified here? It seems like we want a file handler for all loggers.

prattmic:
LGTM

server/server/settings.py
pmtischler: From the [Django docs](https://docs.djangoproject.com/en/1.9/topics/logging/#django):
> django is the catch-all logger. No messages are posted directly to this logger.

My best guess is that my definition of 'django' creates the 'default' logger for all Django components, and 'py.warnings' handles the rest.

PR #46, opened 2016-01-31
Don't install virtualenv-3.4

prattmic:
The latest version (1.11.0) of the stankevich-python Puppet module will
specify the appropriate version of Python to use to the system
virtualenv install. It is no longer necessary to use a separate
virtualenv installation for Python 3.

This is the recommended path forward from pypa/virtualenv#851, as the
versioned virtualenv binaries are likely going away.


pmtischler:
LGTM once tests complete and show passing.

prattmic:
I didn't even try it locally, because Travis is way faster than my system. :P

pmtischler:
:)

Issue #47, opened 2016-02-01
Fix references to rules in the docs


pmtischler:
[Example](http://auvsi-suas-competition-interoperability-system.readthedocs.org/en/latest/spec.html#TargetTypes)

PR #48, opened 2016-02-02
Update rule refs and remove IR target type.

pmtischler:
For #47


prattmic:
LGTM

prattmic:
Can you run `makemigrations` and send another PR? The Target model changed when IR was removed.

pmtischler:
That's confusing, I swear I did that. Done in #58

Issue #45, opened 2016-01-31
Add objects to capture mission clock.


pmtischler:
Enables the interoperability system to judge total mission time.

Enables the future target review system to know when targets can be considered frozen and ready for review.

@prattmic Thoughts before I start?

prattmic:
Nothing in particular. Big buttons for each event (mission start, time out, etc) seems reasonable. Making previous times editable on the same page would be nice too, but lower priority. We could go into the admin interface to edit times.

pmtischler:
Ok. I was thinking the same workflow as TakeoffOrLandingEvents for now, and then make an interface for this in #37.

pmtischler:
Upcoming PRs:
1) Refactor TakeoffOrLandingEvent.flights so generic implementation it is part of TimePeriod.
2) Refactor TakeoffOrLandingEvent.user_in_air so it uses a function in AccessLog which gets the latest access log before a time. 
3) Add a method similar to flights and user_in_air to MissionClockEvent in order to get the on-mission-clock periods, on timeout periods, and whether the user is on mission clock / timeout.
4) Add grading of time-on-mission-clock.

PR #49, opened 2016-02-02
Add MissionClockEvent to track mission clock time.

pmtischler:
For #45 

Following this PR:
1) Refactor [TakeoffOrLandingEvent.flights](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/models/takeoff_or_landing_event.py#L22) so generic implementation it is part of  [TimePeriod](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/models/time_period.py).
2) Refactor [TakeoffOrLandingEvent.user_in_air](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/models/takeoff_or_landing_event.py#L64) so it uses a function in `AccessLog` which gets the latest access log before a time. 
3) Add a method similar to `flights` and `user_in_air` to `MissionClockEvent` in order to get the on-mission-clock periods, on timeout periods, and whether the user is on mission clock / timeout.
4) Add grading of time-on-mission-clock.
5) In future target grading interfaces, hide submitted targets for users which are either on mission clock or on timeout.

### DIFF
--- a/server/auvsi_suas/models/mission_clock_event.py
+++ b/server/auvsi_suas/models/mission_clock_event.py
@@ -1,0 +1,22 @@
+"""Mission Clock event model."""
+
+from access_log import AccessLog
+from time_period import TimePeriod
+from django.db import models
+from django.utils import timezone
+
+
+class MissionClockEvent(AccessLog):
+    """Marker for a team going on/off mission clock."""
+    # Whether the team is now on the mission clock.
+    team_on_clock = models.BooleanField()
+    # Whether the team is now in a timeout.
+    team_on_timeout = models.BooleanField()
+
+    def __unicode__(self):
+        """Descriptive text for use in displays."""
+        return unicode('MissionClockEvent (pk:%s, user:%s, timestamp:%s, '
+                       'team_on_clock:%s, team_on_timeout:%s)' %
+                       (str(self.pk), self.user.__unicode__(),
+                        str(self.timestamp), str(self.team_on_clock),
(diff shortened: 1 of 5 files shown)


pmtischler:
Forgot to run formatter, doing that now.

server/auvsi_suas/models/mission_clock_event.py
```diff
@@ -0,0 +1,22 @@
...
+class MissionClockEvent(AccessLog):
+    """Marker for a team going on/off mission clock."""
+    # Whether the team is now on the mission clock.
+    team_on_clock = models.BooleanField()
+    # Whether the team is now in a timeout.
+    team_on_timeout = models.BooleanField()
+
+    def __unicode__(self):
+        """Descriptive text for use in displays."""
+        return unicode('MissionClockEvent (pk:%s, user:%s, timestamp:%s, '
```
prattmic: No space before ' ('

prattmic:
LGTM

I wonder if this is the best way to approach the problem. All of these edge events are difficult to query and figure out the times when the team is on the clock. I'm not really sure there is a better way to do this, but it seems like what we really want is a list of time ranges when the team is on the clock.

pmtischler:
I thought about that as well. The list of time ranges is an implementation detail of this overall idea, and I think each has their +/-s. If we come up with a clear reason to switch we can always revisit, this part of the code base is small & easy to maintain.

PR #52, opened 2016-02-05
Use namedtuple as base client type. Add id,user to Target.

pmtischler:
For #50 

Our custom serialization/deserialization can be handled by `collections.namedtuple`. This PR refactors to use that instead, which will be useful going forward (easier to add new types and serialize/deserialize in multiple places). Note this makes client types immutable.

Add the id,user fields to Target which were missing in last PR.

### DIFF
--- a/client/interop/types.py
+++ b/client/interop/types.py
@@ -5,85 +5,92 @@ requires. They include input validation, making a best-effort to ensure
 values will be accepted by the server.
 """
 
+from collections import namedtuple
+from collections import OrderedDict
 import dateutil.parser
 import re
 import sys
+
+
+def check_type(obj, obj_type):
+    """Asserts that the obj is of the given type.
+
+    Raises:
+        ValueError: Invalid type.
+    """
+    if not isinstance(obj, obj_type):
+        raise ValueError('Object (%s) is not the type %s' % (obj, obj_type))
+
+
+def check_bounds(value, min_value, max_value):
+    """Assert that the given value is within the range.
+
+    Raises:
+        ValueError: Out of range.
+    """
+    if value < min_value or value > max_value:
+        raise ValueError('Value (%f) out of range [%f, %f]' %
+                         (value, min_value, max_value))
 
 
 def check_latitude(lat):
     """Assert that latitude is valid.
 
     Raises:
-        ValueError: Latitude out of range
+        ValueError: Latitude out of range.
     """
-    if lat < -90 or lat > 90:
-        raise ValueError('Latitude (%f) out of range [-90 - 90]' % lat)
+    check_bounds(lat, -90, 90)
 
 
 def check_longitude(lon):
     """Assert that longitude is valid.
 
     Raises:
-        ValueError: Longitude out of range
+        ValueError: Longitude out of range.
     """
-    if lon < -180 or lon > 180:
-        raise ValueError('Longitude (%f) out of range [-180, 180]' % lon)
+    check_bounds(lon, -180, 180)
 
 
-class Serializable(object):
-    """ Serializable is a simple base class which provides basic
-    'serialization' (a simple dict) of a subset of attributes the inheriting
-    class.
+def check_heading(heading):
+    """Asserts that heading is valid.
 
-    By serializing only specified attributes, other attributes can be utilized
-    by the class (or its subclasses) without being included in the serialized
-    dict.
+    Raises:
+        ValueError: Heading out of range.
     """
-
-    def __init__(self, attrs):
-        """Initialize attributes to serialize
-
-        Args:
-            attrs: List of string attribute names to serialize.
-        """
-        self.attrs = attrs
-
-    def serialize(self):
-        """Serialize the current state of the object."""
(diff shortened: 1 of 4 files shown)


pmtischler:
Sigh. https://hg.python.org/cpython/rev/fa3ac31cfa44

pmtischler:
PTAL.

pmtischler:
Note that `_asdict()` seems to have been reverted in August 2015 ([discussion](http://grokbase.com/t/python/python-bugs-list/158rqgbk20/issue24931-asdict-breaks-when-inheriting-from-a-namedtuple), [commit](https://hg.python.org/cpython/rev/fa3ac31cfa44)). The Ubuntu version is 3.4.3 which was [released in February 2015](https://www.python.org/downloads/) before that fix. We can probably remove our `_asdict()` overrides if we upgrade the version of Python. Thoughts?

prattmic:
Sorry, I thought I replied to this a few days ago.

I'm not sure I understand the motivation behind this change. It seems to generally be a regression in functionality. There were a few reasons for implementing our own type instead of using namedtuple originally:

* A custom type is more pleasant to use directly. namedtuple requires all fields to be passed to the constructor. A normal class can be built up one attribute at a time, as data becomes available. It is also mutable, so it can be easily updated as necessary. Both of these uses break our sanity checks, but they are just sanity checks, not critical, and we could fix that problem by uses @property attributes.
* The custom type is easier to subtype. Teams may well want to subtype something like client.Target to use as their core Target type, which includes extra data they need, but is still directly serializable by the client. This is doable with namedtuple, but as above, it not very nice.

I have a few more comments inline.

client/interop/types.py
```diff
@@ -5,85 +5,92 @@
...
-    def __init__(self, latitude, longitude, altitude_msl, uas_heading):
-        super(Telemetry, self).__init__(['latitude', 'longitude',
-                                         'altitude_msl', 'uas_heading'])
-
-        self.latitude = float(latitude)
-        self.longitude = float(longitude)
-        self.altitude_msl = float(altitude_msl)
-        self.uas_heading = float(uas_heading)
+    def __init__(self, *args, **kwargs):
+        check_type(self.latitude, float)
```
prattmic: Don't do this. Previously we only checked that the fields were castable to float, not float explicitly. Explicitly requiring float makes this type needlessly painful to use.

client/interop/types.py
```diff
@@ -5,85 +5,92 @@
...
+        check_type(self.uas_heading, float)
 
         check_latitude(self.latitude)
         check_longitude(self.longitude)
+        check_heading(self.uas_heading)
 
-        if self.uas_heading < 0 or self.uas_heading > 360:
-            raise ValueError('Heading (%f) out of range [0, 360]' %
-                             self.uas_heading)
+    def _asdict(self):
```
prattmic: Please leave a comment explaining why this is necessary.

pmtischler:
My original intentions were:
1. *Remove custom serialization code and use a standard approach*. This isn't captured by this PR because of the `namedtuple` subclassing bug. `namedtuple` as a choice also created more issues.
2. *Single set of code for serializing/deserializing*. In particular, deserializing is not in the class atm and the result is duplication. Point (1) made it easy by leveraging a base class.
3. *Make it easier to define / edit interfaces*.

Your points:
1. *Incremental object building*. I agree that it is a feature and would like to have it. 
2. *Client side validation*. The more I think about it, I'd prefer to remove client-side validation. It's an unnecessary nice-to-have that increases our work and surface for bugs. Perfect validation is just a LAN round-trip `O(1 ms)` away. All typing and range issues should be resolved in testing before competition. We should be leveraging something from my point (1) to solve any typing issues, rather than us manually adding typing to Python.
3. *Teams subclassing client code*. I hadn't thought about teams wanting to do this. I think the best (safest) way for teams to do this is to have a custom class have the client types as a member, rather than inheriting. Either way we shouldn't use this as a motivating factor in the discussion.
4. *Cast to float vs. Require float*. Agree that it makes it much more painful, but that may be a feature which exposes unexpected bugs in teams' implementations. I'm happy either way though.

My net thought atm is:
1. Revert this PR
2. Remove client-side validation
3. Use some system to standardize our interface definition and do to/from json/python (Maybe [Protobuf Json](https://developers.google.com/protocol-buffers/docs/proto3#json)?)
4. Add the id,user fields 

Thoughts?

prattmic:
I don't feel strongly about client-side validation. I think it is nice to have, but you are right that the real test is to try to do the real thing.

I'm not sure I understand the issue with the custom serialization code. It is literally 2 lines of actual code. Deserialization should be something as simple as:

```python
def deserialize(self, d):
    [self.__dict__[k] = d[k] for k in d if k in self.attrs]
```

I suspect Protobuf JSON would be difficult to use. It is intended to use JSON as a Protobuf serialization format, not as a interface between native JSON and Protobufs. i.e., I don't think you get direct control over the precise fields it creates when serializing.

pmtischler:
#59 created as a replacement to this PR.

pmtischler:
Deleting this PR.

PR #56, opened 2016-02-10
Heading and orientation reference true north

### LINKED ISSUE #29: Specify true north for headings/directions
Interoperability interface does not specify whether angles are relative to true north or magnetic north. We should specify true north in:

* The interopability interface documentation
* Comments / code

Once complete, we should notify teams via the Google Group.

prattmic:
Explicitly state that the UAS telemetry heading and target orientation
are both based on true north, not magnetic north.

Closes #29


pmtischler:
LGTM

Issue #30, opened 2016-01-27
Add ADLC fields to target submission


pmtischler:
The interoperability target submission interface needs whether the team is eligible for ADLC credit or not. We need 3 boolean fields as to whether a given target was automatically detected, classified, and/or localized.

* Update the code
* Update the interface
* Update the docs
* Notify teams via Google Groups

pmtischler:
Any update on this?

prattmic:
A single boolean 'autonomous' field will suffice. Per section 7.3, they get localization credit if the position is within 150ft, and characterization credit if three characteristics are correct. The first six detected targets must be submitted.

pmtischler:
SG

PR #57, opened 2016-02-10
Add 'autonomous' Target field

prattmic:
This field is used distinguish targets that are submitted as ADLC
targets. The rules are strict about only grading the first six
submissions, so these targets are locked down and may not be changed or
deleted once submitted.

References #30

### DIFF
--- a/server/auvsi_suas/views/targets_test.py
+++ b/server/auvsi_suas/views/targets_test.py
@@ -114,6 +114,7 @@ class TestPostTarget(TestCase):
             'alphanumeric': 'ABC',
             'alphanumeric_color': 'black',
             'description': 'Best target',
+            'autonomous': False,
         }
 
         response = self.client.post(targets_url,
@@ -135,6 +136,7 @@ class TestPostTarget(TestCase):
         self.assertEqual(target['alphanumeric_color'],
                          created['alphanumeric_color'])
         self.assertEqual(target['description'], created['description'])
+        self.assertEqual(target['autonomous'], created['autonomous'])
 
         # It also contains 'user' and 'id' fields.
         self.assertIn('id', created)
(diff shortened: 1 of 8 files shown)


pmtischler:
We are discussing this on the email thread. Holding off review for now.

pmtischler:
We've reached consensus on the email thread. We are going to add a field `autonomous` and treat it like the other fields. That is, no additional write time validation. If at end of mission clock the team has >6 ADLC targets submitted, they will get no credit. To implement this we will export the total ADLC targets submitted, and a downstream grading spreadsheet will compute the score appropriately.

We should indicate this on the API, and tell teams when we notify about API change.

pmtischler:
Any update?

prattmic:
I've been quite busy and haven't had a chance to update this. Hopefully
I'll get to it this week.

prattmic:
PTAL

server/auvsi_suas/views/targets_test.py
```diff
@@ -329,6 +332,23 @@ def test_invalid_orientation(self):
                                         content_type='application/json')
             self.assertEqual(400, response.status_code)
 
+    def test_invalid_autonomous(self):
+        """Send bad target autonomous."""
+        bad = ['true', 1, 'Yes']
```
pmtischler: Why is `true` bad? Note that in the docs above it uses `false` rather than `False`.

pmtischler:
LGTM. Please make an announcement on the Google Groups when it is submitted and the new documentation is hosted on readthedocs.

server/auvsi_suas/views/targets_test.py
prattmic: The field requires a JSON boolean, not a string. Here a Python boolean would become a JSON boolean, but a string is just s string.

prattmic:
Done

Issue #44, opened 2016-01-31
Add target types to interoperability client


pmtischler:

pmtischler:
@prattmic 

I started to work on this issue, and I'm wondering what's the best way to handle the enumeration of acceptable types. In my branch I've imported the enums from the server, and I've updated the requirements to reference the server's (though it's not correct atm). This doesn't seem like the best way to do it. Thoughts?
* [types.py](https://github.com/auvsi-suas/interop/blob/target-types/client/interop/types.py)
* [requirements.txt](https://github.com/auvsi-suas/interop/blob/target-types/client/requirements.txt)

prattmic:
No, we shouldn't depend on the server in the client. I think the simplest
solution is to just copy the enums and have duplicates in the server and
client. I would also be OK with moving them to the client and making the
server depend on the client.

pmtischler:
I agree client shouldn't depend on server. I don't think server should depend on client. Duplicating it seems uni-deal from maintenance standpoint.

The client doesn't have full correctness checking for other interfaces- it simply does sanity checks. I think trying to duplicate validation logic on both sides is not worth our time. I think we should just remove the non sanity-checks and let the server respond with an error to notify students of the mistake. Agree?

prattmic:
Regardless of checks, I think it is nice to provide an enum of values for teams to use, but I don't think it is critical or urgent.

pmtischler:
Need to add upload of target image.

PR #59, opened 2016-02-11
Remove client side validation. Add deserialize method.

pmtischler:
Replacement for PR #52 
For #44

### DIFF
--- a/client/interop/types.py
+++ b/client/interop/types.py
@@ -10,26 +10,6 @@ import re
 import sys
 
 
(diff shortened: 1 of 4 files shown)


client/interop/types.py
```diff
@@ -40,18 +20,15 @@ class Serializable(object):
     dict.
```
prattmic: Please note that subtypes must set attrs

prattmic:
LGTM

PR #60, opened 2016-02-13
Add id,user to Target. Change target_type to type.

pmtischler:
Replaces #52. For #44

### DIFF
--- a/client/interop/types.py
+++ b/client/interop/types.py
@@ -176,7 +176,11 @@ class Target(Serializable):
     """A target.
 
     Attributes:
-        target_type: Target type, must be one of TargetType.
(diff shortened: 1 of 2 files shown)


client/interop/types.py
```diff
@@ -194,7 +198,9 @@ class Target(Serializable):
     """
 
     def __init__(self,
-                 target_type,
+                 id=None,
+                 user=None,
+                 type=None,
```
prattmic: Nit: `type` is a builtin. It is preferable not to override it. `self.type` is fine, but the argument should probably have a different name.

prattmic:
LGTM

PR #61, opened 2016-02-13
Refactor time periods from flights. Add time periods to clock.

pmtischler:
For #45

### DIFF
--- a/server/auvsi_suas/models/time_period.py
+++ b/server/auvsi_suas/models/time_period.py
@@ -4,9 +4,51 @@ class TimePeriod(object):
     A value of None indicates infinity. Both start and end are inclusive.
     """
 
+    @classmethod
+    def from_events(cls, events, is_start_func, is_end_func):
+        """Gets a list of time periods from a list of events.
+
+        Args:
+            events: List of events which have a timestamp property.
+            is_start_func: Lambda which returns whether an event is the start
+                of a period.
+            is_end_func: Lambda which returns whether an event is the end
+                of a period.
+        Returns:
+            A list of time periods.
+        """
+        if not events:
+            return []
+
+        periods = []
+        # Handle end with no start, use None as infinity.
+        if events and is_end_func(events[0]):
+            periods.append(TimePeriod(None, events[0].timestamp))
+        # Detect transitions from start->end and end->start.
+        start_time = None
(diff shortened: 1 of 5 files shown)


server/auvsi_suas/models/time_period.py
```diff
@@ -4,10 +4,50 @@ class TimePeriod(object):
...
+                of a period.
+            is_end_func: Lambda which returns whether an event is the end
+                of a period.
+        Returns:
+            A list of time periods.
+        """
+        periods = []
+
+        # Handle end with no start, use None as infinity.
+        if events and is_end_func(events[0]):
```
prattmic: Before this return [] if len(events) < 1

server/auvsi_suas/models/time_period.py
```diff
@@ -4,10 +4,50 @@ class TimePeriod(object):
...
+
+        return periods
+
     def __init__(self, start=None, end=None):
         self.start = start
         self.end = end
 
+    def __unicode__(self):
+        """Descriptive text for use in displays."""
+        return unicode('TimePeriod (start: %s, end: %s)' % (self.start,
```
prattmic: Nit: `TimePeriod(start=%s, end=%s)`

server/auvsi_suas/models/time_period_test.py
```diff
@@ -7,6 +8,62 @@
 class TestTimePeriod(TestCase):
     """Tests the TimePeriod object."""
 
+    def setUp(self):
+        self.Event = namedtuple('Event', ['timestamp', 'start'])
```
prattmic: Why make this a class attribute instead of just a normal global class?

prattmic:
LGTM

server/auvsi_suas/models/time_period.py
pmtischler: Done.
```
if not events:
  return []
```

server/auvsi_suas/models/time_period.py
pmtischler: Done.

server/auvsi_suas/models/time_period_test.py
pmtischler: Done.

PR #63, opened 2016-02-14
Add target endpoints to client.

pmtischler:
For #44

### DIFF
--- a/client/interop/client.py
+++ b/client/interop/client.py
@@ -9,11 +9,12 @@ See README.md for more details."""
 
 from concurrent.futures import ThreadPoolExecutor
 import functools
+import json
 import requests
 import threading
 
 from .exceptions import InteropError
-from .types import ServerInfo, StationaryObstacle, MovingObstacle
+from .types import ServerInfo, StationaryObstacle, MovingObstacle, Target
 
 
 class Client(object):
@@ -29,10 +30,10 @@ class Client(object):
 
         Args:
             url: Base URL of interoperability server
-                (e.g., http://localhost:8000)
-            username: Interoperability username
-            password: Interoperability password
-            timeout: Individual session request timeout (seconds)
+                (e.g., http://localhost:8000).
+            username: Interoperability username.
+            password: Interoperability password.
+            timeout: Individual session request timeout (seconds).
         """
         self.url = url
         self.timeout = timeout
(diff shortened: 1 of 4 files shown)


client/interop/client.py
```diff
@@ -127,6 +153,77 @@ def get_obstacles(self):
...
 
+    def get_targets(self):
+        """GET targets.
+
+        Returns:
+            List of Target objects which are viewable by user.
+        Raises:
+            InteropError: Error from server.
+            requests.Timeout: Request timeout.
+            ValueError or AttributeError: Mlaformed response from server.
```
prattmic: Malformed

prattmic: And several times below

client/interop/client_test.py
```diff
@@ -141,3 +141,55 @@ def test_get_obstacles(self):
...
         self.assertIn(300, async_heights)
         self.assertIn(200, async_heights)
+
+    def test_targets(self):
+        """Test target workflow."""
+        # Initially no targets.
+        targets = self.client.get_targets()
+        async_targets = self.async_client.get_targets().result()
+
+        self.assertEqual([], targets)
```
prattmic: It is probably best to avoid this assertion, since we are running against a live, potentially unclean server. E.g., if one run of this fails halfway through the method, the targets it created won't be cleaned up.

client/interop/types.py
```diff
@@ -10,23 +10,37 @@
...
+        for attr in self.attrs:
+            if self.__dict__[attr] != other.__dict__[attr]:
+                return False
+        return True
+
+    def __str__(self):
+        """Gets string encoding of object."""
+        return "%s (%s)" % (self.__class__.__name__,
+                            ', '.join('%s:%s' % (attr, self.__dict__[attr])
+                                      for attr in self.attrs))
```
prattmic: Make this "%s(%s)" with join ("%s=%s") to make this the canonical Python form. Also, probably rename this method __repr__, since that is what it is.

client/interop/client_test.py
pmtischler: Done.

client/interop/client.py
pmtischler: Done.

client/interop/client_test.py
```diff
@@ -141,3 +141,47 @@ def test_get_obstacles(self):
...
+
+        self.assertEquals(post_target, put_target)
+        self.assertEquals(async_post_target, async_put_target)
+
+        # Delete target.
+        self.client.delete_target(post_target.id)
+        self.async_client.delete_target(async_post_target.id).result()
+
+        targets = self.client.get_targets()
+        self.assertEquals([], targets)
```
prattmic: Same issue here.

pmtischler: Done.

prattmic:
LGTM

Issue #24, opened 2016-01-10
akashcas@ - Python 2/3 tests failing


akashcas:
I git pulled the most recent changes. I then created a new vagrant machine, ran vagrant setup for automated setup . The Python2 and python 3  test failed. [Output](http://pastebin.com/QXnxa5Aq)

prattmic:
Could you please post the contents of `/var/log/apache2/interop_server_access.log` and `/var/log/apache2/interop_server_error.log`? They should contain more details about the server error.

pmtischler:
@akashcas Any update?

akashcas:
@pmtischler @prattmic I overcame that error by doing the complete manual setup using setup.sh . All the test ran successfully for the first time but later i ran the test.sh for the same again it gave me the same error . Sorry late reply as i was out of station.

akashcas:
Why i am getting the same error when i am running the test again after once it showed me all successful in earlier case .

prattmic:
I'm not sure, could you please post the contents of the files I mentioned the previous time you had trouble: `/var/log/apache2/interop_server_access.log` and `/var/log/apache2/interop_server_error.log`?

akashcas:
interop_server_access.log : 

```
127.0.0.1 - - [27/Jan/2016:18:47:30 +0530] "POST /api/login HTTP/1.1" 200 17 "-" "python-requests/2.9.1"
127.0.0.1 - - [27/Jan/2016:18:47:31 +0530] "POST /api/login HTTP/1.1" 200 17 "-" "python-requests/2.9.1"
127.0.0.1 - - [27/Jan/2016:18:47:31 +0530] "GET /api/obstacles HTTP/1.1" 200 196 "-" "python-requests/2.9.1"
127.0.0.1 - - [27/Jan/2016:18:47:31 +0530] "GET /api/obstacles HTTP/1.1" 200 195 "-" "python-requests/2.9.1"
127.0.0.1 - - [27/Jan/2016:18:47:31 +0530] "POST /api/login HTTP/1.1" 200 17 "-" "python-requests/2.9.1"
127.0.0.1 - - [27/Jan/2016:18:47:31 +0530] "POST /api/login HTTP/1.1" 200 17 "-" "python-requests/2.9.1"
127.0.0.1 - - [27/Jan/2016:18:47:31 +0530] "GET /api/server_info HTTP/1.1" 200 128 "-" "python-requests/2.9.1"
[...]

PR #65, opened 2016-02-15
Perform Vagrant up in Travis.

pmtischler:
For #24


prattmic:
Travis runs tests in a VM which does not support nested virtualization. Vagrant will not work.

pmtischler:
Doh. Ok Removing.

PR #67, opened 2016-02-28
Fix middleware for non-processed requests.

pmtischler:


prattmic:
LGTM, though it would be nice for the log to say "unknown time" instead of -1s.

pmtischler:
Done.

PR #68, opened 2016-02-28
Fix bug in evaluate_teams dedupe logs.

pmtischler:


prattmic:
How did this pass tests before?

pmtischler:
Afk atm, but it would need more than one log to be exposed.

prattmic:
We should try to add regression tests for this.

pmtischler:
Yes we should. I checked the data and we do have multiple telemetry posts per user, which is odd.

pmtischler:
Figured it out. If it returns only 1 flight period then it won't do the duplicate check either.

I added a regression test by adding an additional flight period, and validated it would have caught it. PTAL

prattmic:
LGTM

PR #76, opened 2016-03-22
Add grading of mission clock.

pmtischler:
For #45

### DIFF
--- a/server/auvsi_suas/models/mission_config.py
+++ b/server/auvsi_suas/models/mission_config.py
@@ -11,6 +11,7 @@ from auvsi_suas.patches.simplekml_patch import AltitudeMode
 from auvsi_suas.models import units
 from fly_zone import FlyZone
 from gps_position import GpsPosition
+from mission_clock_event import MissionClockEvent
 from moving_obstacle import MovingObstacle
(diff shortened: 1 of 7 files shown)


server/auvsi_suas/models/mission_config.py
```diff
@@ -164,6 +166,15 @@ def evaluate_teams(self):
             # Start the evaluation data structure.
             eval_data = results.setdefault(user, {})
 
+            # Calculate the total mission clock time.
+            mission_clock_time = 0
+            missions = MissionClockEvent.missions(user)
+            for mission in missions:
+                duration = mission.duration()
+                if duration is not None:
```
prattmic: None means infinite, probably because we forgot to update the clock, right? I feel like we should have a way of denoting/warning about that.

prattmic:
LGTM besides the note about infinite time.

server/auvsi_suas/models/mission_config.py
pmtischler: I've added a list of warnings to the evaluation. It gets returned as a column in the CSV. I also added a warning for unbounded flight periods.

I think that function is getting really messy. I'm planning to refactor and clean it up once the target work is complete.

PR #80, opened 2016-03-26
Add target image requests to client.

### LINKED ISSUE #44: Add target types to interoperability client

pmtischler:
Fixes #44


client/interop/client.py
```diff
@@ -219,11 +219,47 @@ def delete_target(self, target_id):
...
+            target_id: The ID of the target for which to get the image.
+        Returns:
+            The image data that was previously uploaded.
+        Raises:
+            InteropError: Error from server.
+            requests.Timeout: Request timeout.
+        """
+        return self.get('/api/targets/%d/image' % target_id).content
+
+    def put_target_image(self, target_id, image_data):
```
prattmic: We should also provide post_target_image. It can just call put_target_image.

client/interop/client_test.py
```diff
@@ -179,6 +179,28 @@ def test_targets(self):
         self.assertEquals(post_target, put_target)
         self.assertEquals(async_post_target, async_put_target)
 
+        # Upload target image.
+        with open('/interop/client/testdata/A.jpg', 'rb') as f:
```
prattmic: More general:

`os.path.join(os.path.dirname(__file__), "../testdata/A,jpg")`

prattmic:
LGTM

client/interop/client.py
pmtischler: Done.

client/interop/client_test.py
pmtischler: Done. That's much better :)

Issue #54, opened 2016-02-09
Admin target review


pmtischler:
This tool will allow judges to review imagery submitted by teams to verify image matches submitted target details.

pmtischler:
@prattmic Do you think we should extend the existing `targets` API with special cases for admin users? Or separate out into a separate interface? I think the two are very different and should be separate, but I don't feel strongly and wanted to check with you. Specifically, we need an interface which returns all targets and their details, and an interface which updates whether they are approved or not.

prattmic:
I think it depends on the differences. In general, I think if there are big differences, we should probably separate them, but I'm not sure how big the differences are.

For instance, GET /api/targets already returns a list of targets for the current user, and GET /api/telemetry allows admins to get telemetry from and (or all) users. It would make sense to extend GET /api/target similarly.

Target approval should perhaps be a different interface (`/api/targets/{id}/approval` or something). I don't think that even superusers should be able to (accidentally) change users targets through POST /api/targets/{id}, so I am reluctant to put an approval field there.

pmtischler:
Agreed about `/api/targets/id/approval`.

For getting the targets I believe there will be a significant enough change, without reuse beyond the `target.json()` call. We will be returning all targets across all users which have completed their mission, rather than returning the targets for the current user. Maybe `/api/targets/review`?

prattmic:
SGTM

pmtischler:
We should update the [user check](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/views/targets.py#L194) to allow superusers to view data. The new interface will return target info for all reviewable targets and then accept updates for whether targets pass review, but we'll need to also grab the image using the current interface.

pmtischler:
Status Update: The review interface is almost complete. I need to finish connecting the approve/deny buttons to the AJAX requests, and then finish the remaining tests.

PR #82, opened 2016-03-26
Add field for judges to approve thumbnails for grading.

pmtischler:
For #54


prattmic:
LGTM, but we should be careful. Nothing stops a team from changing the thumbnail after we set that field while they still have access.

pmtischler:
Ack. My intention is to have the judge grade interface only operate on
targets for teams off mission clock, so they shouldn't be able to edit it
at that point.

Issue #83, opened 2016-03-26
Finish documentation for MissionPlanner integration


pmtischler:
[TODO(jemoster)](https://github.com/auvsi-suas/interop/blob/master/docs/integration/mission_planner.rst)

pmtischler:
@prattmic Can you take this?

prattmic:
I've never actually used Mission Planner, so I may not be best for this.

pmtischler:
Neither have I :(

PR #86, opened 2016-04-01
Add basic Mission Planner docs

prattmic:
These aren't very detailed, but they are a quick start.

Related to #83


pmtischler:
LGTM after Travis failure fix

PR #90, opened 2016-04-04
Add target upload tool.

### LINKED ISSUE #53: Admin target upload
Teams are allowed to submit via tab delimited file submitted via USB drive. This interface will allow judges to add/edit those targets via interoperability server admin interface. Targets will then be auto-evaluated.

pmtischler:
Fixes #53 

Added MEDIA_URL as well so that in debug mode you can view the target images in the admin display. Note that Django doesn't let MEDIA_URL run in non-debug mode (no security on image access).

### DIFF
--- a/client/targets/upload_targets.py
+++ b/client/targets/upload_targets.py
@@ -1,0 +1,169 @@
+#!/usr/bin/env python
+# Program to load targets from file and upload via interoperability.
+
+import argparse
+import concurrent
+import csv
+import getpass
+import imghdr
+import os
+import re
+
+from interop import Client
+from interop import Target
+
+TARGET_TYPE_MAP = {
+    'STD': 'standard',
+    'QRC': 'qrc',
+    'OAX': 'off_axis',
+    'EMG': 'emergent',
+}
+
+LATITUDE_REGEX = re.compile(
+    '(?P<dir>[NS])(?P<deg>\d\d) (?P<min>\d\d) (?P<sec>\d\d\.\d{0,3})')
+LATITUDE_DIR = {'S': -1, 'N': 1}
+LONGITUDE_REGEX = re.compile(
+    '(?P<dir>[EW])(?P<deg>\d\d\d) (?P<min>\d\d) (?P<sec>\d\d\.\d{0,3})')
+LONGITUDE_DIR = {'E': -1, 'W': 1}
+
+
+def load_target_file(target_filepath):
+    """Loads targets from the given file.
+
+    Args:
+        target_filepath: The path to the target file to load.
+    Returns:
+        A list of (target, image_filepath) tuples.
+    Raises:
+        ValueError if the file is not properly formatted.
+    """
+    targets = []
+    with open(target_filepath, 'r') as f:
+        reader = csv.reader(f, delimiter='\t')
+        for row in reader:
+            target_type = row[1]
+            latitude_str = row[2]
+            longitude_str = row[3]
+            orientation = row[4].lower()
+            shape = row[5]
+            background_color = row[6]
+            alphanumeric = row[7]
+            alphanumeric_color = row[8]
+            image_filepath = row[9]
+            description = row[10]
+
+            target = Target()
+
+            if target_type not in TARGET_TYPE_MAP:
+                raise ValueError('Type %s not in %s' % (target_type,
+                                                        str(TARGET_TYPE_MAP)))
+            target.type = TARGET_TYPE_MAP[target_type]
+
+            # Parse latitude. Not required for off axis.
+            if target.type != 'off_axis':
+                match = LATITUDE_REGEX.match(latitude_str)
+                if match:
+                    latitude = LATITUDE_DIR[match.group('dir')] * (
+                        float(match.group('deg')) + float(match.group('min')) /
+                        60 + float(match.group('sec')) / 3600)
+                    target.latitude = latitude
+                else:
+                    raise ValueError('Latitude is not valid: %s' %
+                                     latitude_str)
+
+            # Parse longitude. Not required for off axis.
+            if target.type != 'off_axis':
+                match = LONGITUDE_REGEX.match(longitude_str)
+                if match:
+                    longitude = LONGITUDE_DIR[match.group('dir')] * (
+                        float(match.group('deg')) + float(match.group('min')) /
+                        60 + float(match.group('sec')) / 3600)
+                    target.longitude = longitude
+                else:
+                    raise ValueError('Longitude is not valid: %s' %
+                                     longitude_str)
+
+            if orientation:
+                target.orientation = orientation
+            if shape:
+                target.shape = shape
+            if background_color:
+                target.background_color = background_color
+            if alphanumeric:
+                target.alphanumeric = alphanumeric
+            if alphanumeric_color:
+                target.alphanumeric_color = alphanumeric_color
+            if description:
+                target.description = description
+            targets.append((target, image_filepath))
+    return targets
+
+
+def main(url, username, password, target_filepath, imagery_dir):
+    """Program to load targets and upload via interoperability.
+
+    Args:
+        url: The interoperability URL.
+        username: The interoperability username.
+        password: The interoperability password.
+        target_filepath: Filepath to the tab-delimited target file.
+        imagery_dir: Base to form paths to imagery files.
+    """
+    # Load target details.
+    targets = load_target_file(target_filepath)
+
+    # Form full imagery filepaths.
+    targets = [(t, os.path.join(imagery_dir, i)) for t, i in targets]
+    # Validate filepath for each image.
+    for _, image_filepath in targets:
+        if not os.path.exists(image_filepath):
+            raise ValueError('Could not find imagery file: %s' %
+                             image_filepath)
+    # Validate type of each image.
+    for _, image_filepath in targets:
+        image_type = imghdr.what(image_filepath)
+        if image_type not in ['jpeg', 'png']:
+            raise ValueError('Invalid imagery type: %s' % image_type)
+
+    # Create client and upload targets.
+    client = Client(url, username, password)
+    for target, image_filepath in targets:
+        image_data = None
+        with open(image_filepath, 'rb') as f:
+            image_data = f.read()
+        target = client.post_target(target)
+        client.put_target_image(target.id, image_data)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='AUVSI SUAS Target Upload.')
+    parser.add_argument('--url',
+                        required=True,
+                        help='URL for interoperability.')
+    parser.add_argument('--username',
+                        required=True,
+                        help='Username for interoperability.')
+    parser.add_argument(
+        '--password',
+        help=
+        'Passowrd for interoperability. If not provided, it will be requested.')
+    parser.add_argument('--target_filepath',
+                        required=True,
+                        help='Filepath to target file.')
+    parser.add_argument(
+        '--imagery_dir',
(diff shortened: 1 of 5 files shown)


client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
+#! /usr/bin/env python
```
prattmic: Not: Space before `/usr/bin/env` seems odd to me.

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+            background_color = row[6]
+            alphanumeric = row[7]
+            alphanumeric_color = row[8]
+            description = row[10]
+
+            target = Target()
+
+            if target_type not in TARGET_TYPE_MAP:
+                print 'Type %s not in %s' % (target_type, str(TARGET_TYPE_MAP))
+                sys.exit(-1)
```
prattmic: `raise ValueError("Type %s not ...")`

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+            shape = row[5]
+            background_color = row[6]
+            alphanumeric = row[7]
+            alphanumeric_color = row[8]
+            description = row[10]
+
+            target = Target()
+
+            if target_type not in TARGET_TYPE_MAP:
+                print 'Type %s not in %s' % (target_type, str(TARGET_TYPE_MAP))
```
prattmic: We should spend the minimal effort to make sure this works on Python 2 and Python 3.

```py
from __future__ import print_function
print(...)
```

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+from interop import Target
+
+TARGET_TYPE_MAP = {
+    'STD': 'standard',
+    'QRC': 'qrc',
+    'OAX': 'off_axis',
+    'EMG': 'emergent',
+}
+
+LATITUDE_REGEX = re.compile('([NS])(\d\d) (\d\d) (\d\d\.\d{0,3})')
```
prattmic: Optional: consider named groups `(?P<name>...)`

https://docs.python.org/2/library/re.html

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+                sys.exit(-1)
+            target.type = TARGET_TYPE_MAP[target_type]
+
+            match = LATITUDE_REGEX.match(latitude_str)
+            if match:
+                latitude = LATITUDE_DIR[match.group(1)] * (
+                    float(match.group(2)) + float(match.group(3)) / 60 +
+                    float(match.group(4)) / 3600)
+                target.latitude = latitude
+            elif target.type != 'off_axis':
```
prattmic: This is unintuitive. At minimum, leave a comment, but I think it would be cleaner to move this up:

```py
# Latitude not needed for off axis targets
if target.type != 'off_axis':
    match = LATITUDE_REGEX.match()
```

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+                target.shape = shape
+            if background_color:
+                target.background_color = background_color
+            if alphanumeric:
+                target.alphanumeric = alphanumeric
+            if alphanumeric_color:
+                target.alphanumeric_color = alphanumeric_color
+            if description:
+                target.description = description
+            image_filepath = row[9]
```
prattmic: Why not put this with the rest of the row parsing?

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='AUVSI SUAS Target Upload.')
+    parser.add_argument('--url',
+                        required=True,
+                        help='URL for interoperability.')
+    parser.add_argument('--username',
+                        required=True,
+                        help='Username for interoperability.')
+    parser.add_argument('--target_filepath',
```
prattmic: Optional: this could be the positional argument.

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+                        required=True,
+                        help='Filepath to target file.')
+    parser.add_argument('--imagery_base_filepath',
```
prattmic: --imagery_dir is shorter and IMO just as clear.

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+                        required=True,
+                        help='Base filepath appended to paths in target file.')
```
prattmic: prepended

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+    args = parser.parse_args()
+
+    if not os.path.exists(args.target_filepath):
+        print 'Target file provided does not exist: %s' % args.target_filepath
+        sys.exit(-1)
+
+    password = getpass.getpass('Interoperability Password: ')
```
prattmic: Optional: this is nice, but for scripting it would be nice to have an argument too.

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+        username: The interoperability username.
+        password: The interoperability password.
+        target_filepath: Filepath to the tab-delimited target file.
+        imagery_base_filepath: Base to form paths to imagery files.
+    """
+    # Load target details.
+    targets = load_target_file(target_filepath)
+
+    # Form full imagery filepaths.
+    targets = [(t, imagery_base_filepath + i) for t, i in targets]
```
prattmic: `os.path.join(imagery_base_filepath, i)`

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+            print 'Invalid imagery type: %s' % image_type
+            sys.exit(-1)
+
+    # Create client and upload targets.
+    client = AsyncClient(url, username, password)
+    target_futures = []
+    for target, image_filepath in targets:
+        target_future = client.post_target(target)
+        target_future.original_target = target
+        target_future.image_filepath = image_filepath
```
prattmic: Maybe I've been using statically typed languages for too long, but this feels like an egregious hack.

client/targets/upload_targets.py
```diff
@@ -0,0 +1,182 @@
...
+            sys.exit(-1)
+    # Validate type of each image.
+    for _, image_filepath in targets:
+        image_type = imghdr.what(image_filepath)
+        if image_type not in ['jpeg', 'png']:
+            print 'Invalid imagery type: %s' % image_type
+            sys.exit(-1)
+
+    # Create client and upload targets.
+    client = AsyncClient(url, username, password)
```
prattmic: I feel like the AsyncClient is complicating things significantly. None of this is latency sensitive and the images will be small. This could be simplified significantly by using a normal Client and doing everything serially.

client/testdata/targets.txt
```diff
@@ -0,0 +1,5 @@
+01	STD	N30 35 34.123	W075 48 47.123	N	rectangle	red	A	orange	A.jpg	AutoDetect
```
prattmic: This may be a bad example, because the tool doesn't tag this as autonomous.

server/auvsi_suas/views/urls.py
```diff
@@ -41,5 +41,5 @@
     url(r'^auvsi_admin/live.kml$', LiveKml.as_view(), name='live_kml'),
     url(r'^auvsi_admin/update.kml$', LiveKmlUpdate.as_view(),
         name='update_kml'),
-) + static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
+) + static(settings.STATIC_URL, document_root=settings.STATIC_ROOT) + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
prattmic: So this does nothing in non-debug mode? How does that work?

pmtischler: Correct. The media URL handles user-based media, and Django doesn't want to make the assumption that user-based media should be publicly visible by default. It'll work in debug mode, but in non-debug you need to provide your own handler so that proper authentication is done.

client/targets/upload_targets.py
pmtischler: Done

client/targets/upload_targets.py
pmtischler: Done

client/targets/upload_targets.py
pmtischler: No more print statements left

client/targets/upload_targets.py
pmtischler: Done

client/targets/upload_targets.py
pmtischler: Done.

client/targets/upload_targets.py
pmtischler: Ack. I prefer it this way- everything is required so it seems arbitrary to make one a positional. I also prefer explicit args to positional unless there is a natural ordering (e.g. with `cp` command)

client/targets/upload_targets.py
pmtischler: Done.

client/targets/upload_targets.py
pmtischler: Done.

client/targets/upload_targets.py
pmtischler: Done.

client/targets/upload_targets.py
pmtischler: No longer there after AsyncClient removal

client/targets/upload_targets.py
pmtischler: It was, it's ridiculously simpler now. I was interested in how you could do it with a pure async flow :)

client/testdata/targets.txt
pmtischler: Done (this was taken from the rules, which doesn't have an actual spec for ADLC)

client/targets/upload_targets.py
pmtischler: Done.

client/targets/upload_targets.py
pmtischler: Done. Note this means that if a user provides a latitude for an off axis target (who know why they would), it will be ignored

client/targets/upload_targets.py
pmtischler: And fwiw, I prefer Angular's async flow with promises to this.

server/auvsi_suas/views/urls.py
prattmic: I still don't understand how this works. MEDIA_URL and MEDIA_ROOT are just variables. Does Django clear them if you aren't in debug mode or something? static() can't know what was passed.

client/targets/upload_targets.py
```diff
@@ -0,0 +1,167 @@
...
+LONGITUDE_DIR = {'E': -1, 'W': 1}
+
+
+def load_target_file(target_filepath):
+    """Loads targets from the given file.
+
+    Args:
+        target_filepath: The path to the target file to load.
+    Returns:
+        A list of (target, image_filepath) tuples.
```
prattmic: Raises: ...

prattmic:
LGTM besides the few extra comments.

server/auvsi_suas/views/urls.py
pmtischler: Not sure how it works either, but it's documented behavior and I validated it.

client/targets/upload_targets.py
pmtischler: Done.

PR #92, opened 2016-04-07
Note that Client handles authentication cookie

### LINKED ISSUE #91: Clarification on Login Cookie
Hello,

The documentation states that the POST /api/login endpoint returns a session id or cookie (stored as Set-Cookie in the response) that must be resent to verify all future requests. The documentation also suggests preparing for errors that involve logging in again in case of failure. 

However, the login view only returns Django's standard HttpResponse with a message (line 51, login.py). My team has tried resetting the server and changing attributes of the user; the only thing required for a 200 request is a previous login, without storing any cookie. 

Does our client need to worry about a cookie from the POST /api/login endpoint?

Thank you,
- Matt T / RU Autonomous

prattmic:
Fixes #91

### DIFF
--- a/client/interop/client.py
+++ b/client/interop/client.py
@@ -20,6 +20,9 @@ from .types import ServerInfo, StationaryObstacle, MovingObstacle, Target
 class Client(object):
     """Client which provides authenticated access to interop API.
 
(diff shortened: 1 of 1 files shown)


pmtischler:
LGTM

Issue #94, opened 2016-04-07
Travis NPM fetch errors


prattmic:
I think this may be the issue: https://github.com/npm/npm/issues/12196

Builds are failing like:

```
npm ERR! registry error parsing json
npm ERR! SyntaxError: Unexpected token 
npm ERR! ï¿½ï¿½Sï¿½nï¿½0ï¿½ï¿½caSï¿½ï¿½Ä¶ï¿½ï¿½ï¿½Iï¿½ï¿½(ï¿½$ï¿½#ï¿½Iï¿½U
ï¿½{ï¿½ï¿½ï¿½ï¿½]ï¿½tD<ß½ï¿½ï¿½taZ(`ï¿½ï¿½ï¿½d'ï¿½ï¿½&dï¿½wï¿½ï¿½ï¿½qï¿½ï¿½Yqa~ï¿½Cyï¿½ï¿½>
ï¿½Kï¿½"ï¿½iï¿½ï¿½Kï¿½ï¿½
npm ERR! ï¿½,ï¿½ï¿½P~]R'-ï¿½>5pJtï¿½"4ï¿½ï¿½ï¿½ï¿½ï¿½C?ï¿½ï¿½gc+ÇŠï¿½YXgï¿½ï¿½T
ï¿½}_21ï¿½ï¿½ï¿½ jï¿½ï¿½ï¿½uï¿½ï¿½tï¿½,Fï¿½/T
J`Pï¿½ï¿½ï¿½Iï¿½*ï¿½ï¿½R$5=ï¿½
[...]

PR #95, opened 2016-04-08
Use puppet 4.x via puppetlab's repo. Install npm via puppet.

pmtischler:
For #94

### DIFF
--- a/server/auvsi_suas/static/auvsi_suas/test_with_phantomjs.sh
+++ b/server/auvsi_suas/static/auvsi_suas/test_with_phantomjs.sh
@@ -1,2 +1,2 @@
-export PHANTOMJS_BIN=/usr/local/bin/phantomjs
+export PHANTOMJS_BIN=/usr/lib/node_modules/phantomjs/bin/phantomjs
 karma start auvsi_suas.conf.karma.js --browsers PhantomJS --single-run
--- a/setup/setup.sh
+++ b/setup/setup.sh
@@ -8,6 +8,7 @@ set -e
 
 SETUP=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
 REPO=$(readlink -f ${SETUP}/..)
+PUPPET=/opt/puppetlabs/bin/puppet
 
 # $1 = log message
 function log() {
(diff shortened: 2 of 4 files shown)


pmtischler:
Note this also resolves some other puppet errors I've seen in the past.

setup/setup.sh
```diff
@@ -41,6 +42,11 @@ exec &> >(tee ${SETUP}/setup-$(date +%F-%H-%M-%S).log)
 log "Creating softlinks..."
 sudo ln -snf ${REPO} /interop
 
+# Install repos.
+log "Installing repos..."
+wget https://apt.puppetlabs.com/puppetlabs-release-pc1-precise.deb -O /tmp/puppetlabs-release-pc1-precise.deb
```
prattmic: Does this deb add a repo to sources.list to keep puppet up to date? If so, can we do a minimum version check here rather than downloading the whole thing every time this script is run?

prattmic:
Will this successfully upgrade a system that has already been setup, or will it fail with apt or puppet errors?

setup/setup.sh
pmtischler: Afaik, yes. How do you recommend doing the minimum version check?

prattmic: I discovered dpkg-query from http://askubuntu.com/a/712202

Warning, this is tested only with the bash interpreter in my head. :)

```sh
version=$(dpkg-query --showformat='${Version}' --show puppet)
major_version=${version%%.*}
if [[ $? != 0 || (( ${major_version} < 4 )) ]]; then
    # wget puppet
    # dpkg install
fi
```

pmtischler: Done.

pmtischler:
It needed an extra flag when removing the previous `npm`, but it now works for both. I've tested with my personal machine.

server/auvsi_suas/static/auvsi_suas/test_with_phantomjs.sh
```diff
@@ -1,2 +1,2 @@
 export PHANTOMJS_BIN=/usr/local/bin/phantomjs
-karma start auvsi_suas.conf.karma.js --browsers PhantomJS --single-run
+sudo karma start auvsi_suas.conf.karma.js --browsers PhantomJS --single-run
```
prattmic: I don't like this. It is running some JS tests, there is no reason it should require root. Some permissions somewhere must be wrong.

pmtischler: Fixed. Turns out I had to update the phantomjs bin path.

prattmic:
LGTM. The push build failure is fixed by https://github.com/auvsi-suas/interop/pull/88

Issue #55, opened 2016-02-09
Target auto-evaluator


pmtischler:
Automatically evaluate targets submitted by a team against a set of correct targets. Dependent on #53 and #54. This data will be exported with the evaluation CSV.

pmtischler:
Clarification that all targets submitted while in air will be judged for actionable intelligence.

pmtischler:
For auto evaluation, the matching value for a submitted target and a real target will be
```
def match_value(geotag_accuracy, num_classify_correct):
  if num_classify_correct < 2:
    return 0
  classify_value = 2 if num_classify_correct == 5 else 1
  geotag_value = 0
  if geotag_accuracy < 75:
    geotag_value = 4
  elif geotag_accuracy < 150:
    geotag_value = 2
  return classify_value + geotag_value
```

pmtischler:
@prattmic Actionable intelligence will use 2 for objective, 1 for threshold. Per discussion with Kim, we will need to update these constants at the competition.

Issue #31, opened 2016-01-27
Add submission time for targets, use for Actionable Intelligence


pmtischler:
Targets which are submitted during the first flight period should get actionable intelligence credit. If teams update the target after the first flight period, they won't get credit for actionable intelligence. To handle this we need to add a timestamp for the submission.

* Update code
* Update interop interface & docs
* Notify teams

pmtischler:
We should have a timestamp for first created, and for last updated.

pmtischler:
See #55 for how we will judge actionable intelligence.

pmtischler:
I believe this is complete? You added the create/modified times.

prattmic:
The code is done, I should update the docs.

pmtischler:
Actually, do you want to create a PR to auto-grade actionable, or should I?

prattmic:
I can do this.

pmtischler:
With #118, I believe this is done.

PR #100, opened 2016-04-11
Add automatic evaluation of submitted targets.

pmtischler:
For #55

Note, after #31 we will have to add in a field for actionable.

### DIFF
--- a/server/auvsi_suas/migrations/0016_missionconfig_targets.py
+++ b/server/auvsi_suas/migrations/0016_missionconfig_targets.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [('auvsi_suas', '0015_target_thumbnail_approved'), ]
+
+    operations = [
+        migrations.AddField(
+            model_name='missionconfig',
+            name='targets',
+            field=models.ManyToManyField(related_name='missionconfig_targets',
+                                         to='auvsi_suas.Target'), ),
+    ]
--- a/server/auvsi_suas/models/mission_config.py
+++ b/server/auvsi_suas/models/mission_config.py
@@ -18,6 +18,8 @@ from server_info import ServerInfo
 from server_info_access_log import ServerInfoAccessLog
 from stationary_obstacle import StationaryObstacle
 from takeoff_or_landing_event import TakeoffOrLandingEvent
+from target import Target
+from target import TargetEvaluator
 from time_period import TimePeriod
 from uas_telemetry import UasTelemetry
 from waypoint import Waypoint
@@ -47,6 +49,10 @@ class MissionConfig(models.Model):
     search_grid_points = models.ManyToManyField(
         Waypoint,
         related_name='missionconfig_search_grid_points')
+
+    # The judge created targets.
+    targets = models.ManyToManyField(Target,
+                                     related_name='missionconfig_targets')
 
     # The last known position of the emergent target
     emergent_last_known_pos = models.ForeignKey(
@@ -136,6 +142,7 @@ class MissionConfig(models.Model):
                     id: Boolean,
                 }
                 'out_of_bounds_time': Seconds spent out of bounds,
+                'targets': Data from TargetEvaluation,
                 'interop_times': {
                     'server_info': {'max': Value, 'avg': Value},
                     'obst_info': {'max': Value, 'avg': Value},
@@ -211,6 +218,11 @@ class MissionConfig(models.Model):
                 out_of_bounds_time += FlyZone.out_of_bounds(
                     self.fly_zones.all(), logs)
             eval_data['out_of_bounds_time'] = out_of_bounds_time
+
+            # Evaluate the targets.
+            user_targets = Target.objects.filter(user=user).all()[:100]
+            evaluator = TargetEvaluator(user_targets, self.targets.all())
+            eval_data['targets'] = evaluator.evaluation_dict()
 
             # Determine interop rates.
             interop_times = eval_data.setdefault('interop_times', {})
--- a/server/auvsi_suas/models/target.py
+++ b/server/auvsi_suas/models/target.py
@@ -1,9 +1,11 @@
 """Target model."""
 
 import enum
-from gps_position import GpsPosition
+import collections
+import networkx as nx
 from django.conf import settings
 from django.db import models
+from gps_position import GpsPosition
 
 
 class Choices(enum.IntEnum):
@@ -185,7 +187,7 @@ class Target(models.Model):
                            thumbnail_approved=str(self.thumbnail_approved),
                            **d))
 
-    def json(self):
+    def json(self, is_superuser=False):
         """Target as dict, for JSON."""
         target_type = None
         if self.target_type is not None:
@@ -221,7 +223,7 @@ class Target(models.Model):
         if self.description != '':
             description = self.description
 
-        return {
+        d = {
             'id': self.pk,
             'user': self.user.pk,
             'type': target_type,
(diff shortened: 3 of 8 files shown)


pmtischler:
This isn't yet ready for full review as I need to add tests.

@prattmic Can you take a look and let me know if the general approach LGs? If so I'll start writing tests.

server/auvsi_suas/models/target.py
```diff
@@ -235,3 +237,151 @@ def json(self):
...
+        # Compute the full matching
+        return nx.bipartite.maximum_matching(g)
+
+    def evaluation_dict(self):
+        """Gets the evaluation dictionary.
+
+        Returns:
+            A dictionary of the form:
+            {
+                'matched_target_value': Total value from matched targets
```
pmtischler: Note: These 2 derived fields will be used in the actual score sheet.

server/auvsi_suas/models/target.py
```diff
@@ -235,3 +237,151 @@ def json(self):
...
+                'unmatched_target_count': Count of unmatched targets
+                'targets': {
```
pmtischler: Note: This dictionary will be provided as feedback for teams to get insight into how they performed, but isn't necessary for scoring.

server/auvsi_suas/models/target.py
```diff
@@ -235,3 +237,149 @@ def json(self):
...
+        classify_fields = {
+            TargetType.standard: standard_fields,
+            TargetType.qrc: ['description'],
+            TargetType.off_axis: standard_fields,
+            TargetType.emergent: [],
+        }
+        fields = classify_fields[self.type]
+        count = 0
+        for field in fields:
+            if getattr(self, field) == getattr(other, field):
```
prattmic: self.field == other.field?

What is getattr adding here?

prattmic: derp, nvm

prattmic:
Yeah, the approach seems reasonable.

pmtischler:
I've added `target` module tests and initial `mission_config` module tests. I need to update the `sample_mission` dataset and add additional tests in `mission_config`. I think this is ready to review, though. PTAL

server/auvsi_suas/models/mission_config.py
```diff
@@ -212,6 +219,18 @@ def evaluate_teams(self):
                     self.fly_zones.all(), logs)
             eval_data['out_of_bounds_time'] = out_of_bounds_time
 
+            # Evaluate the targets.
+            eval_data['targets'] = {}
+            user_targets = Target.objects.filter(user=user).all()
+            target_sets = {
+                'human': [t for t in user_targets if not t.autonomous],
```
pmtischler: Note I've broken it down by human/auto as they are evaluated separately.

server/auvsi_suas/models/target.py
```diff
@@ -235,3 +238,153 @@ def json(self):
...
+                        'location_accuracy': distance from actual
+                    }
+                }
+            }
+        """
+        matched_target_value = 0
+        unmatched_target_count = 0
+        target_dict = collections.defaultdict(dict)
+
+        for submitted in self.submitted_targets:
```
pmtischler: Actually, we need to do this by real target so that the column headings are consistent per team, and so they get feedback for targets they didn't submit as well.

server/auvsi_suas/migrations/0016_missionconfig_targets.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [('auvsi_suas', '0015_target_thumbnail_approved'), ]
```
prattmic: This needs to be updated now.

Sorry from taking so long to review...

server/auvsi_suas/models/mission_config.py
prattmic: Nit: I've always used the terms manual vs auto

server/auvsi_suas/models/target.py
```diff
@@ -235,3 +238,153 @@ def json(self):
...
+            TargetType.standard: standard_fields,
+            TargetType.qrc: ['description'],
+            TargetType.off_axis: standard_fields,
+            TargetType.emergent: [],
+        }
+        fields = classify_fields[self.target_type]
+        count = 0
+        for field in fields:
+            if getattr(self, field) == getattr(other, field):
+                count += 1
```
prattmic: If both self and other are missing the field, it will count as a match. Is that desired?

server/auvsi_suas/models/target.py
```diff
@@ -235,3 +238,153 @@ def json(self):
...
+        # Targets which don't have an approved image have no match value.
+        if (submitted.target_type != real.target_type or
+                not submitted.thumbnail_approved):
+            return 0
+
+        # Compute the classification point value.
+        i = bisect.bisect_right(settings.TARGET_CLASSIFY_VALUE,
+                                (real.similar_classifications(submitted),
+                                 float('inf')))
+        classify_value = settings.TARGET_CLASSIFY_VALUE[i - 1][1] if i else 0
```
prattmic: This is all very clever, but I think it would be a lot more readable if you just used ranges that were checked explicitly.

```py
TARGET_CLASSIFY_RANGES = [
     {"start": float("-inf"), "end": 2./5, "value": 0},
     {"start": 2. / 5, "end": 1, "value": 1}
     {"start": 1, "end": float("inf"), "value": 2}
 ]

def classify_value(similar):
    for r in TARGET_CLASSIFY_RANGE:
        if similar >= r["start"] and similar < r["end"]:
            return r["value"]

    raise AssertionError("Incomplete TARGET_CLASSIFY_RANGE")
```

server/auvsi_suas/models/target.py
pmtischler: Other should be a judge created target listed in the `MissionConfig`, so it shouldn't be missing.

server/auvsi_suas/models/mission_config.py
pmtischler: Done.

server/auvsi_suas/models/target.py
pmtischler: Fixed.

server/auvsi_suas/migrations/0016_missionconfig_targets.py
pmtischler: Done.

server/auvsi_suas/models/target.py
pmtischler: Done.

pmtischler:
This is ready for review. PTAL

server/auvsi_suas/models/target.py
```diff
@@ -247,3 +250,174 @@ def json(self):
             'description': description,
             'autonomous': self.autonomous,
         }
+        if is_superuser:
```
prattmic: I believe this conflicts with your other PR?

server/auvsi_suas/models/target.py
```diff
@@ -247,3 +250,174 @@ def json(self):
...
+            submitted = self.matches.get(real)
+            if submitted:
+                submitted_pk = submitted.pk
+                match_value = self.match_value(submitted, real)
+                thumbnail_approved = submitted.thumbnail_approved
+                classifications = submitted.similar_classifications(real)
+                location_accuracy = submitted.location.distance_to(
+                    real.location)
+            matched_target_value += match_value
+            target_dict[real.pk] = {
```
prattmic: Perhaps we should just omit targets with no submitted match? Just a thought.

prattmic:
LGTM

server/auvsi_suas/models/target.py
pmtischler: Reverted this part.

server/auvsi_suas/models/target.py
pmtischler: Done.

PR #102, opened 2016-04-14
Drop interop client dependency from docs

prattmic:
ReadTheDocs runs `pip install` from the root of the repo and doesn't
seem to be configurable, so it can't properly install the interop
client.

Instead of adding a dependency, teach Sphinx to mock the client
dependencies and import the module directly.


pmtischler:
LGTM.

prattmic:
PTAL

pmtischler:
LGTM

PR #107, opened 2016-04-21
Make puppet exit with a non-zero code on failure

prattmic:
For some absurd reason, puppet exits with a zero exit code, even if the
configuration fails to apply. --detailed-exitcodes is needed to make it
exit non-zero:
https://docs.puppet.com/puppet/latest/reference/man/agent.html#OPTIONS

Now Travis builds should fail early when the configuration fails to
apply.

### DIFF
--- a/setup/setup.sh
+++ b/setup/setup.sh
@@ -76,6 +76,16 @@ ensure_puppet_module stankevich-python
 
 # Launch the Puppet process. Prepares machine.
 log "Executing Puppet setup..."
+set +e
 sudo ${PUPPET} apply \
+    --detailed-exitcodes \
     --modulepath=${SETUP}/puppet_files:/etc/puppetlabs/code/environments/production/modules \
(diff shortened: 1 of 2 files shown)


setup/setup.sh
```diff
@@ -76,6 +76,16 @@ ensure_puppet_module stankevich-python
 
 # Launch the Puppet process. Prepares machine.
 log "Executing Puppet setup..."
+set +e
 sudo ${PUPPET} apply \
+    --detailed-exitcodes \
     --modulepath=${SETUP}/puppet_files:/etc/puppetlabs/code/environments/production/modules \
     ${SETUP}/puppet_files/auvsi_suas.pp
+
+# Exit codes 0 and 2 indicate success, so rewrite code 2 as 0.
```
pmtischler: Well aint that annoying.

pmtischler:
LGTM

PR #108, opened 2016-04-21
Add creation and last modified time to Target

prattmic:
References #31

### DIFF
--- a/server/auvsi_suas/models/target.py
+++ b/server/auvsi_suas/models/target.py
@@ -164,6 +164,14 @@ class Target(models.Model):
     # Whether judge considers thumbnail valid for target.
     thumbnail_approved = models.NullBooleanField()
 
+    # Time that this target was first created.
+    # auto_now_add automatically sets this field on initial creation.
+    creation_time = models.DateTimeField(auto_now_add=True)
+
+    # Time that this target was last modified.
+    # auto_now automatically updates this field on save().
+    last_modified_time = models.DateTimeField(auto_now=True)
+
     def __unicode__(self):
         """Descriptive text for use in displays."""
         d = self.json()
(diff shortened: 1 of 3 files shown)


server/auvsi_suas/models/target.py
```diff
@@ -164,6 +164,18 @@ class Target(models.Model):
     # Whether judge considers thumbnail valid for target.
     thumbnail_approved = models.NullBooleanField()
 
+    # Time that this target was first created.
+    # auto_now_add automatically sets this field on initial creation.
+    # null allows existing rows to have a null value, making migrations
+    # feasible. New rows will all have a value.
+    creation_time = models.DateTimeField(auto_now_add=True, null=True)
```
pmtischler: I think you can specify `null` is valid in the migration, but not valid here, so it will migrate old data but not allow new `null` data.

server/auvsi_suas/models/target.py
```diff
@@ -164,6 +164,18 @@ class Target(models.Model):
     # Whether judge considers thumbnail valid for target.
     thumbnail_approved = models.NullBooleanField()
 
+    # Time that this target was first created.
+    # auto_now_add automatically sets this field on initial creation.
+    # null allows existing rows to have a null value, making migrations
```
pmtischler: You don't need this comment (null allows ...)- this is standard in the repo at this point.

pmtischler:
LGTM

server/auvsi_suas/models/target.py
prattmic: done

server/auvsi_suas/models/target.py
prattmic: gone

PR #111, opened 2016-05-03
Provide more details to admins.

pmtischler:
For #54

### DIFF
--- a/server/auvsi_suas/models/target.py
+++ b/server/auvsi_suas/models/target.py
@@ -177,30 +177,12 @@ class Target(models.Model):
 
     def __unicode__(self):
         """Descriptive text for use in displays."""
-        d = self.json()
+        d = self.json(is_superuser=True)
+        return unicode('{name}({fields})'.format(
+            name=self.__class__.__name__,
+            fields=', '.join('%s=%s' % (k, v) for k, v in d.iteritems())))
 
-        return unicode("{name}(pk={id}, user={user}, "
-                       "target_type={type}, "
-                       "latitude={latitude}, "
-                       "longitude={longitude}, "
-                       "orientation={orientation}, shape={shape} "
-                       "background_color={background_color}, "
-                       "alphanumeric='{alphanumeric}', "
-                       "alphanumeric_color={alphanumeric_color}, "
-                       "description='{description}', "
-                       "autonomous='{autonomous}', "
-                       "thumbnail='{thumbnail}', "
-                       "thumbnail_approved='{thumbnail_approved}', "
-                       "creation_time={creation_time}, "
-                       "last_modified_time={last_modified_time})".format(
-                           name=self.__class__.__name__,
-                           thumbnail=self.thumbnail,
-                           thumbnail_approved=str(self.thumbnail_approved),
-                           creation_time=self.creation_time,
-                           last_modified_time=self.last_modified_time,
-                           **d))
-
-    def json(self):
+    def json(self, is_superuser=False):
         """Target as dict, for JSON."""
         target_type = None
         if self.target_type is not None:
@@ -236,7 +218,7 @@ class Target(models.Model):
         if self.description != '':
             description = self.description
 
-        return {
+        d = {
             'id': self.pk,
             'user': self.user.pk,
             'type': target_type,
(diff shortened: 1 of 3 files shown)


server/auvsi_suas/models/target.py
```diff
@@ -174,30 +174,10 @@ class Target(models.Model):
...
-                           name=self.__class__.__name__,
-                           thumbnail=self.thumbnail,
-                           thumbnail_approved=str(self.thumbnail_approved),
-                           creation_time=self.creation_time,
-                           last_modified_time=self.last_modified_time,
-                           **d))
-
-    def json(self):
+        d = self.json(is_admin=True)
+        return unicode(', '.join('%s=%s' % (k, v) for k, v in d.iteritems()))
```
prattmic: Don't you want a class name on the front? Or at least some kind of surrounding brackets?

server/auvsi_suas/models/target.py
```diff
@@ -247,3 +227,19 @@ def json(self):
             'description': description,
             'autonomous': self.autonomous,
         }
+
+        if is_admin:
+            thumbnail = None
+            if self.thumbnail is not None:
+                thumbnail = self.thumbnail
+            d['thumbnail'] = thumbnail
```
prattmic: What am I missing? Why not `d['thumbnail'] = self.thumbnail`?

server/auvsi_suas/models/target_test.py
```diff
@@ -135,13 +135,22 @@ def test_json(self):
...
         self.assertEqual(True, d['autonomous'])
+        self.assertNotIn('thumbnail_approved', d)
+
+        d = t.json(is_admin=True)
+        self.assertIn('thumbnail_approved', d)
+
+        t.thumbnail_approved = True
+        t.save()
+        d = t.json(is_admin=True)
+        self.assertEqual(True, d['thumbnail_approved'])
```
prattmic: Check for None thumbnail?

server/auvsi_suas/models/target.py
```diff
@@ -174,30 +174,10 @@ class Target(models.Model):
...
-                           thumbnail_approved=str(self.thumbnail_approved),
-                           creation_time=self.creation_time,
-                           last_modified_time=self.last_modified_time,
-                           **d))
-
-    def json(self):
+        d = self.json(is_admin=True)
+        return unicode(', '.join('%s=%s' % (k, v) for k, v in d.iteritems()))
+
+    def json(self, is_admin):
```
prattmic: why not default is_admin to False?

server/auvsi_suas/views/targets.py
```diff
@@ -172,7 +172,7 @@ def post(self, request):
                    autonomous=data.get('autonomous', False))
         t.save()
 
-        return JsonResponse(t.json(), status=201)
+        return JsonResponse(t.json(request.user.is_superuser), status=201)
```
prattmic: this is inconsistent to the above, which uses the kwarg.

server/auvsi_suas/views/targets.py
```diff
@@ -212,7 +212,7 @@ def get(self, request, pk):
         except ValueError as e:
             return HttpResponseForbidden(str(e))
 
-        return JsonResponse(target.json())
+        return JsonResponse(target.json(request.user.is_superuser))
 
     def put(self, request, pk):
```
prattmic: Are we sure we want to allow superusers to modify other users' targets? Perhaps we should log which user last modified a target.

prattmic:
A bunch of small comments, otherwise LGTM

server/auvsi_suas/models/target.py
pmtischler: Ooops! Good catch.

server/auvsi_suas/models/target.py
pmtischler: Done.

server/auvsi_suas/models/target.py
pmtischler: Done.

server/auvsi_suas/models/target_test.py
pmtischler: Done.

server/auvsi_suas/views/targets.py
pmtischler: Done.

server/auvsi_suas/views/targets.py
pmtischler: A superuser already can via the Django admin interface in an unlogged fashion.

On the other hand, I've decided to make a more narrow interfaces anyway for the judging interface, so I've reverted.

PR #112, opened 2016-05-05
Use path field rather than manually form.

pmtischler:


prattmic:
Is use of the path attribute sanctioned? I see it in the Django source for FieldFile, but it is not listed in the official docs.

Otherwise LGTM

pmtischler:
[Managing files](https://docs.djangoproject.com/en/1.9/topics/files/) documentation shows the use of `field.path`. I believe it's sanctioned.

PR #113, opened 2016-05-06
Allow admins ability to read targets+images.

pmtischler:
Turns out for review we need an interface for requesting imagery. Rather than duplicate it, give admins access.


prattmic:
LGTM

PR #115, opened 2016-05-11
Add interface to get and update target review.

pmtischler:


pmtischler:
@prattmic While I'm writing tests, what do you think of this API / approach?

server/auvsi_suas/views/targets.py
```diff
@@ -118,13 +119,8 @@ def dispatch(self, *args, **kwargs):
         return super(Targets, self).dispatch(*args, **kwargs)
 
     def get(self, request):
-        # We only support getting up to 100 targets for now.
```
prattmic: I think we should still have a limit to avoid excessive slowdown if there are tons of targets.

prattmic:
The API looks fine to me

server/auvsi_suas/views/targets.py
pmtischler: I see. Reverted & updated the comment.

pmtischler:
Tests added. PTAL

server/auvsi_suas/views/targets_test.py
```diff
@@ -817,3 +820,117 @@ def test_get_after_delete(self):
...
+        self.assertEqual('qrc', data[0]['type'])
+
+    def test_put_review_no_approved(self):
+        """Test PUT review with no approved field."""
+        target = Target(user=self.team, target_type=TargetType.qrc)
+        target.save()
+
+        response = self.client.put(targets_review_id_url(args=[target.pk]))
+        self.assertEqual(400, response.status_code)
+        # TODO
```
prattmic: TODO?

prattmic:
LGTM

server/auvsi_suas/views/targets_test.py
```diff
@@ -817,3 +820,117 @@ def test_get_after_delete(self):
...
+        self.user = User.objects.create_superuser(
+            'testuser', 'testemail@x.com', 'testpass')
+        self.team = User.objects.create_user('testuser2', 'testemail@x.com',
+                                             'testpass')
+        response = self.client.post(login_url, {
+            'username': 'testuser',
+            'password': 'testpass'
+        })
+        self.assertEqual(200, response.status_code)
+        self.maxDiff = None
```
prattmic: This is unused

server/auvsi_suas/views/targets_test.py
pmtischler: Lingering. Removed.

PR #118, opened 2016-05-12
Actionable Intelligence scoring

prattmic:
Score objective/threshold level of actionable intelligence and include
the result in the evaluation dict. No scores are used at this point,
simply an objective/threshold string.

sample_mission.json is unchanged expect for the addition of real and
submitted targets. Unfortunately the database dump reordered the file.

References #31

### DIFF
--- a/server/auvsi_suas/models/target.py
+++ b/server/auvsi_suas/models/target.py
@@ -7,6 +7,7 @@ import operator
 from django.conf import settings
 from django.db import models
 from gps_position import GpsPosition
+from takeoff_or_landing_event import TakeoffOrLandingEvent
 
 
 class Choices(enum.IntEnum):
@@ -268,6 +269,30 @@ class Target(models.Model):
                 count += 1
         return float(count) / len(fields) if fields else 1
 
+    def actionable_submission(self, flights=None):
+        """Checks if Target meets Actionable Intelligence submission criteria.
+
+        A target is "actionable" if the aircraft was in flight from initial
+        target submission until the last update. Note that this does not check
+        the localization or characterization Actionable Intelligence criteria.
+
+        Args:
+            flights: Optional memoized flights for this target's user. If
+                     omitted, the flights will be looked up.
+
+        Returns:
+            True if target may be considered an "actionable" submission.
+        """
+        if flights is None:
+            flights = TakeoffOrLandingEvent.flights(self.user)
+
+        for flight in flights:
+            if flight.within(self.creation_time) and \
+                flight.within(self.last_modified_time):
+                return True
+
+        return False
+
 
 class TargetEvaluator(object):
     """Evaluates submitted targets against real judge-made targets."""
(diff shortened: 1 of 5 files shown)


server/auvsi_suas/models/target.py
```diff
@@ -268,6 +269,25 @@ def similar_classifications(self, other):
...
+        """Checks if Target meets Actionable Intelligence submission criteria.
+
+        A target is "actionable" if the aircraft was in flight from initial
+        target submission until the last update. Note that this does not check
+        the localization or characterization Actionable Intelligence criteria.
+
+        Returns:
+            True if target may be considered an "actionable" submission.
+        """
+        flights = TakeoffOrLandingEvent.flights(self.user)
```
pmtischler: Do we want to pass this in and drop it from `O(targets)` queries to `O(1)`?

pmtischler:
Overall LGTM. Note that we may need to change the match value to include this info based on our offline email thread.

pmtischler:
Per the issue thread, let's add it to the match value and use 2,1 as point values (in settings file) for the scoring.

server/auvsi_suas/models/target.py
prattmic: Done

prattmic:
Done. PTAL

server/auvsi_suas/models/target_test.py
```diff
@@ -327,22 +377,29 @@ def setUp(self):
...
         """Tests the match value for two targets."""
-        e = TargetEvaluator([], [])
-        self.assertEqual(6, e.match_value(self.submit1, self.real1))
+        e = TargetEvaluator(self.submitted_targets, self.real_targets)
+        self.assertEqual(8, e.match_value(self.submit1, self.real1))
         self.assertEqual(3, e.match_value(self.submit2, self.real2))
         self.assertEqual(0, e.match_value(self.submit3, self.real3))
         self.assertEqual(0, e.match_value(self.submit4, self.real4))
 
     def test_match_targets(self):
```
pmtischler: Please add a test case for where real is an empty list, or submitted is an empty list.

server/server/settings.py
```diff
@@ -209,3 +209,17 @@
      "end": float("inf"),
      "value": 0},
 ]
+
+# Actionable Intelligence threshold and objective level parameters.
+TARGET_ACTIONABLE_PARAMS = {
+    "threshold": {
+        "characteristics": 3. / 5.,
```
pmtischler: Would it be better to be semantic instead (say threshold level actionable requires threshold level characteristics & location), rather than via value? That way the two will stay in sync. It may require restructuring the original structure to use named values.

pmtischler:
LGTM after comments.

server/server/settings.py
prattmic: Imagery threshold requires only two characteristics, but Actionable intelligence threshold requires three characteristics.

pmtischler: :0 good catch. Sounds good to do it this way.

server/auvsi_suas/models/target_test.py
prattmic: Done

Issue #117, opened 2016-05-11
No Target Submissions Yields to Bipartite Error


pmtischler:
[match_targets](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/models/target.py#L339) should check for no submitted / real targets, as otherwise it results in a 500 error due to:
```
NetworkXError: Graph is not bipartite.
```

pmtischler:
Added test coverage in #120 to demonstrate it isn't due to lack of targets on either side. Instead, I root caused this to having a MissionConfig reference targets which were submitted by a user, instead of creating a separate judge target. This causes a self reference when matching submitted to real targets, which causes the graph to not be bipartite. This is a configuration error, not a bug.

PR #120, opened 2016-05-15
Improve test coverage for targets.

pmtischler:
Demonstrates #117 isn't due to lack of targets on either side.


pmtischler:
Odd. Travis is reporting failure, but nothing about that test has changed, and everything passes locally.

prattmic:
Can you hold off on this PR? My #118 adds targets to sample_mission.json which can be used for this.

pmtischler:
Actually I'll just revert given #118 adds more coverage. I only created the PR because I had added tests to verify the problem in #117.

PR #121, opened 2016-05-16
Update and refactor prober.

pmtischler:
Use the Interoperability Client instead of the older one-off client.

Merge the two prober mains into a single main with a data generation
selection flag.

Add LatLon as a requirement for the client, which was missing.

Both were tested against the current server code.


client/tools/prober.py
```diff
@@ -0,0 +1,114 @@
...
+        data_generator: The generator of UAS telemetry.
+        interop_time: The time between interoperability execution.
+        logger: The logger to write status to.
+    """
+    # Continually execute interop requests until signaled to stop
+    while True:
+        # Get start time of the round
+        start_time = datetime.datetime.now()
+        # Get generated data
+        data = data_generator.get_uas_telemetry(start_time)
```
prattmic: The generator could return Telemetry directly

prattmic:
\o/ deleted code.  LGTM

client/tools/prober.py
pmtischler: Done.

PR #125, opened 2016-05-20
Add a frontend for target review.

### LINKED ISSUE #54: Admin target review

pmtischler:
Updates the backend review PUT to use the same field so it is directly
updatable by the Angular resource object.

Updates the menus for better organization of displays/tools. Updates the
help message on the dashboard.

Adds a new resource for target review requests. Adds a page to view
and update targets under review.

Fixes #54

### DIFF
--- a/server/auvsi_suas/static/auvsi_suas/pages/target-review/target-review-controller.js
+++ b/server/auvsi_suas/static/auvsi_suas/pages/target-review/target-review-controller.js
@@ -1,0 +1,182 @@
+/**
+ * Controller for the Target Review page.
+ */
+
+
+
+/**
+ * Controller for the Target Review page.
+ * @param {!angular.Window} $window The window service.
+ * @param {!Object} Backend The backend service.
+ * @final
+ * @constructor
+ * @struct
+ * @ngInject
+ */
+TargetReviewCtrl = function($window, Backend) {
+    /**
+     * @private @const {!angular.Window} The window service.
+     */
+    this.window_ = $window;
+
+    /**
+     * @private @const {!Object} The backend service.
+     */
+    this.backend_ = Backend;
+
+    /**
+     * @private {?Object} The target under review.
+     */
+    this.target_ = null;
+
+    /**
+     * @private {?Array<Object>} The target review details.
+     */
+    this.targetDetails_ = null;
+};
+
+
+/**
+ * Gets the targets for review.
+ * @return {?Array<Object>} The targets.
+ * @export
+ */
+TargetReviewCtrl.prototype.getReviewTargets = function() {
+    return this.backend_.reviewTargets;
+};
+
+
+/**
+ * Gets the class for the target.
+ * @param {!Object} target The target.
+ * @return {string} The class.
+ * @export
+ */
+TargetReviewCtrl.prototype.getTargetButtonClass = function(target) {
+    var thumbnailClass;
+    if (target.thumbnail_approved == null) {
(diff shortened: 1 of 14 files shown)


pmtischler:
@prattmic I shared on Drive a demo video.

server/auvsi_suas/static/auvsi_suas/pages/target-review/target-review-controller.js
```diff
@@ -0,0 +1,182 @@
...
+    }
+};
+
+
+/**
+ * Sets the target under review.
+ * @param {!Object} target The target to review.
+ * @export
+ */
+TargetReviewCtrl.prototype.setReviewTarget = function(target) {
```
prattmic: Shouldn't there be some indication of which team the target is associated with?

server/auvsi_suas/static/auvsi_suas/pages/target-review/target-review.html
```diff
@@ -0,0 +1,38 @@
...
+
+    <hr></hr>
+
+    <div class="row text-center text-middle">
+        <div id="target-list" class="button-group stacked">
+            <button type="button"
+                class="target-list-item {{targetReviewCtrl.getTargetButtonClass(target)}}"
+                ng-repeat="target in targetReviewCtrl.getReviewTargets()"
+                ng-click="targetReviewCtrl.setReviewTarget(target)">
+                {{target.id}}
```
prattmic: It would be nice to have more info on the screen than just target IDs. Maybe a grid of thumbnails?

prattmic:
LGTM. I left a couple of usability comments, but they don't need to be addressed now.

server/auvsi_suas/static/auvsi_suas/pages/target-review/target-review-controller.js
pmtischler: Why would that matter? 

This interface is used to verify that for standard/off-axis/qrc targets it is a reasonable image (e.g. not an irrelevant picture), and to validate the emergent text is correct (e.g. "it's a firefighter putting out a fire"). Neither require knowledge of which team it came from.

prattmic: It isn't required, but we are going to want an easy way to verify if we are getting targets from all the different teams.

server/auvsi_suas/static/auvsi_suas/pages/target-review/target-review.html
pmtischler: I suspect that we will just review them in order and never look back (other than an accidental click, for which you would just go back a few). For this use case I don't know the thumbnails will be useful.

I could see us leveraging thumbnails under certain scenarios.We need some extra infrastructure and/or more complex frontend to properly do this. I don't think I have time for such upgrades right now, but feel free to file a feature request for post-competition.

pmtischler:
Added some responses. I'll await your response.

server/auvsi_suas/static/auvsi_suas/pages/target-review/target-review-controller.js
pmtischler: We should get new targets every time a team goes off mission clock, which would provide indication. We can also see this in the Django admin view listing targets.

prattmic:
LGTM

PR #126, opened 2016-05-21
Review targets with image. Review in edit order.

### LINKED ISSUE #123: Target review should return intuitive stable order
It should return targets in order of going uneditable (user going off mission clock), and then by target ID.

pmtischler:
Filter targets which don't have a thumbnail stored. Sort the final list
of targets by last edit time. This will provide a stable order for
reviewing.

Fixes #123


prattmic:
LGTM

PR #127, opened 2016-05-29
Require waypoint capture to be in order

### LINKED ISSUE #71: Waypoint track distance and in-order traversal
We need to evaluate in-order traversal and within the max track distance. The way this should be handled:

* Convert sorted list of telemetry into sets which represent continuous sections within the max track distance. O(n)
* For each set, find which waypoints were hit and only count hits that occur in order. O(n)

prattmic:
Rather than allowing waypoint capture at any point in the logs, verify
that waypoints are captured one-after-another.

For testing, move the second waypoint in sample_mission.json and make
only user1 hit it.

Fixes #71


server/auvsi_suas/views/auvsi_admin/export_kml_test.py
```diff
@@ -92,7 +92,7 @@ def __init__(self, *args, **kwargs):
             for lat, lon, alt in [
                 (-76.0, 38.0, 0.0), (-76.0, 38.0, 10.0), (-76.0, 38.0, 20.0), (
```
pmtischler: This formatting isn't the best.

pmtischler:
LGTM except it doesn't fix #71. It handles threshold waypoint navigation. For objective waypoint navigation we need to add checks for leaving the waypoint path.

server/auvsi_suas/views/auvsi_admin/export_kml_test.py
prattmic: Done

prattmic:
Oops, I thought that requirement was dropped altogether, not moved to objective. I'll make another PR for staying within the track.

Issue #34, opened 2016-01-27
Cleanup tuple lists in tests


pmtischler:
Some of the older tests still have large, difficult to read tables. They should be refactored to be more clear.

pmtischler:
I think we'll tackle this as we make changes. No need to keep an open issue- we agree on the principle.

PR #128, opened 2016-05-30
Verify UAS remains along waypoint track

### LINKED ISSUE #71: Waypoint track distance and in-order traversal

prattmic:
When evaluating waypoints, verify that every UAS telemetry point remains
within 100ft of the straight-line waypoint track.

To achieve this, we convert WGS84 coordinates into a cartesian
coordinate system (UTM), where we can do simply vector arithmetic to
determine the distance from the waypoint track line segment.

Fixes #71

### DIFF
--- a/docs/management/configure_manage.rst
+++ b/docs/management/configure_manage.rst
@@ -146,7 +146,8 @@ The user must create a MissionConfig in order to test interoperability.
    mission. The active mission is used when responding to requests or storing
    data. The admin must make sure there is only one labeled active at a time.
 #. `Home pos` defines the home position. This is the center of the admin
-   display and should be where the ground station is.
+   display and should be where the ground station is. The mission is expected
+   to occur reasonably close to this position (i.e., within several miles).
 #. `Fly zones` define the valid areas where a UAS is considered in-bounds. This
    must include the takeoff and landing strip. This is used to evaluate time
    spent out of bounds.
--- a/server/auvsi_suas/models/distance.py
+++ b/server/auvsi_suas/models/distance.py
@@ -1,7 +1,12 @@
 """Functions for computing distance."""
 
 import math
+import numpy as np
+import pyproj
+from django.conf import settings
 from auvsi_suas.models import units
+
+wgs84 = pyproj.Proj(init="epsg:4326")
 
 
 def haversine(lon1, lat1, lon2, lat2):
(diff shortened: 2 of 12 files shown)


server/auvsi_suas/models/distance_test.py
```diff
@@ -55,3 +56,63 @@ def test_competition_distances(self):
...
+    utm = pyproj.Proj(init="epsg:32618")
+
+    def evaluate_inputs(self, inputs):
+        """Evaluates a list of test cases."""
+        for start, end, point, expected in inputs:
+            dist = distance.distance_to_line(start,
+                                             end,
+                                             point,
+                                             utm_proj=self.utm)
+            self.assertAlmostEqual(expected, dist, delta=5)  # ft
```
pmtischler: 5ft sounds like a lot?

server/auvsi_suas/models/distance_test.py
```diff
@@ -55,3 +56,63 @@ def test_competition_distances(self):
...
+
+    def test_at_points(self):
+        """Test inputs that are at or near the end points."""
+        self.evaluate_inputs([
+            # (start,        end,            point,          dist)
+            ((38, -76, 100), (38, -77, 100), (38, -76, 100), 0),
```
pmtischler: We created #34 to deal with these types of lists. Maybe the shear ease of doing it this way is worth the uncleanliness. How did you plan to do this differently? Is is worth making new tests conform with #34? Or should we close that issue?

server/server/settings.py
```diff
@@ -178,9 +178,18 @@
 # (1.5x safety factor, 10Hz, 4 interfaces)
 TEST_LOADTEST_INTEROP_MIN_RATE = 1.5 * 10.0 * 4
 
+# Universal Transverse Mercator projection zone for the region used by
+# missions on this configuration. This should be a spatial reference accepted
+# by proj. Default is UTM zone 18N, covering Webster Field, MD:
+# http://www.spatialreference.org/ref/epsg/wgs-84-utm-zone-18n/
+UTM_PROJECTION = 'epsg:32618'
```
pmtischler: Instead of hard-coding a value, could we instead infer the zone with a GPS position, and use the mission's `home` position for the projection? That is how the frontend does a projection, and why I included the `home` data point.

setup/puppet_files/auvsi_suas/manifests/server_install.pp
```diff
@@ -7,6 +7,7 @@
 
     # Install packages from apt.
     $package_deps = [
+        'python-pyproj',
```
pmtischler: Won't this be installed by pip?

server/auvsi_suas/models/distance_test.py
```diff
@@ -55,3 +56,63 @@ def test_competition_distances(self):
 
 
 # TODO: Add additional tests for distance_to()
```
pmtischler: I think we can remove this TODO

docs/management/configure_manage.rst
```diff
@@ -162,6 +162,14 @@ The user must create a MissionConfig in order to test interoperability.
 #. Once the MissionConfig is fully defined, click Save.
 
 
+.. note::
+
+    By default, the server is configured to evaluate missions created around
```
pmtischler: I'd prefer teams don't have to edit the settings file. See other comment about inferring the WGS zone from a single position.

server/server/settings.py
pmtischler: [Underlying project docs](https://github.com/OSGeo/proj.4/wiki/GenParms) seem to indicate you can define an origin instead?

server/auvsi_suas/models/mission_config.py
```diff
@@ -117,24 +118,51 @@ def satisfied_waypoints(self, uas_telemetry_logs):
         Args:
             uas_telemetry_logs: A list of UAS Telemetry logs.
         Returns:
-            A list of booleans where each value indicates whether the UAS
-            satisfied the waypoint for that index.
+            satisifed: A list of booleans where each value indicates whether
```
pmtischler: It seems like this variable could be a count now that in-order traversal is required.

server/auvsi_suas/models/mission_config.py
```diff
@@ -117,24 +118,51 @@ def satisfied_waypoints(self, uas_telemetry_logs):
...
+                       the UAS satisfied the waypoint for that index.
+            within_track: A boolean indicating whether the UAS remained along
```
pmtischler: Similar to above, would we want to make this a count as well and just grade on whether the count is all waypoints? This will provide greater feedback to teams, and won't require modification for next year (I hope to change it from all-or-nothing to points per waypoint).

pmtischler:
In general it LGs.

server/server/settings.py
prattmic: The haversine function used in distance_to and in the frontend is not a projection. It is a special case formula for computing the distance between two points on a sphere. To find the closest distance between a point and a line, I need to perform general vector arithmetic, which the haversine function is not going to be sufficient for.

The arithmetic could potentially be converted to a spherical coordinate system, but a) it is much simpler in a cartesian coordinate system, and b) UTM is (AFAIK) by far the most common coordinate system in which most geographic/GIS computations are performed in.

I'm fairly certain that proj cannot automatically select the appropriate UTM projection for us, since it supports converting between hundreds of different projections, but it is not too complicated to manually infer the correct zone, which I will do.

server/auvsi_suas/models/distance_test.py
prattmic: I think this table test is fine (though it is reaching the limit of complexity I like). It is the super complex nested ones I dislike a lot, like: https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/models/moving_obstacle_test.py#L66

server/auvsi_suas/models/distance_test.py
prattmic: Indeed. Unfortunately, I don't have a great way to independently verify these calculations. The 2D distances were calculated with a distance measuring tool and eye-balling the right point on a map, and the 3D distances were the same with some extra trig on my part to compensate for altitude.

setup/puppet_files/auvsi_suas/manifests/server_install.pp
prattmic: Since it uses the C API, installing with pip still requires that python-dev is preinstalled. Since it is not, I figured I may as well simply install it directly from apt anyways, since that is faster.

docs/management/configure_manage.rst
prattmic: Done

server/auvsi_suas/models/mission_config.py
prattmic: Done

server/auvsi_suas/models/mission_config.py
prattmic: Done

server/server/settings.py
prattmic: Done

server/auvsi_suas/models/distance_test.py
prattmic: Well, it technically hasn't been done.

prattmic:
PTAL

server/auvsi_suas/models/mission_config.py
```diff
@@ -117,24 +118,55 @@ def satisfied_waypoints(self, uas_telemetry_logs):
         Args:
             uas_telemetry_logs: A list of UAS Telemetry logs.
         Returns:
-            A list of booleans where each value indicates whether the UAS
-            satisfied the waypoint for that index.
+            satisifed: Total number of waypoints hit (must be in order).
+                       the UAS satisfied the waypoint for that index.
```
pmtischler: This second line seems left over.

server/auvsi_suas/models/mission_config.py
```diff
@@ -117,24 +118,55 @@ def satisfied_waypoints(self, uas_telemetry_logs):
...
+        prev, curr = -1, 0
+        satisfied = 0
+        satisfied_track = 0
+        track_ok = True
 
         for uas_log in uas_telemetry_logs:
             if curr == len(waypoints):
                 break
 
+            # Once we have passed the first waypoint, ensure that the UAS
```
pmtischler: So this only accounts for the first waypoint track attempt. We need to evaluate any attempt (e.g. after a circle around).

setup/puppet_files/auvsi_suas/manifests/server_install.pp
pmtischler: Ok.

server/auvsi_suas/models/distance.py
```diff
@@ -53,3 +58,122 @@ def distance_to(latitude_1, longitude_1, altitude_1, latitude_2, longitude_2,
     gps_dist_ft = units.kilometers_to_feet(gps_dist_km)
     alt_dist_ft = abs(altitude_1 - altitude_2)
     return math.hypot(gps_dist_ft, alt_dist_ft)
+
+
+def utm_zone(lat, lon):
```
pmtischler: I like it :)

pmtischler:
1 new comment regarding multiple attempts. The other stuff LGs.

server/auvsi_suas/models/mission_config.py
prattmic: Done

server/auvsi_suas/models/mission_config.py
prattmic: I am really off my game with this PR...

Done.

prattmic:
PTAL

pmtischler:
LGTM

PR #134, opened 2016-06-17
Drop wrap content starting with text/html

prattmic:
New versions of Django give HTML responses with the content type
'text/html; charset: utf-8'. We don't want to wrap that, so only check
for a start of text/html.


pmtischler:
LG

Issue #139, opened 2016-06-22
Frontend map display has performance problem


pmtischler:
One of the backend requests pegs CPU. Still need to investigate.

pmtischler:
Odd. Running it today didn't produce any problems.

prattmic:
Try sending 10Hz telemetry while using the display?

pmtischler:
I did, I think it's an endurance thing.

prattmic:
I can reproduce it easily.

Loading `/api/teams?debug` shows that the request took 26s, with 131 SQL queries totalling 6.6s waiting on the database.  Many of the queries took 200ms alone.

My VM is obviously less powerful than your machine, but I should be able to do some good profiling.

prattmic:
I think I know how to solve the fundamental issue.

There are four key SQL queries performed for each user, which come from [user_json](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/views/teams.py#L15):

* `MissionClockEvent.user_on_clock(user)`: ~300ms
* `MissionClockEvent.user_on_timeout(user)`: ~300ms
* `TakeoffOrLandingEvent.user_in_air(user)`: ~300ms
* `UasTelemetry.user_active(user)`: ~1ms

The fundamental problem is that [AccessLog](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/models/access_log.py#L11), which MissionClockEvent, TakeoffOrLandingEvent, and UasTelemetry all inherit from, is a first-class model. This means that Django creates a dedicated database table for AccessLog (`auvsi_suas_accesslog`). All of the different subtypes of AccessLog store the the AccessLog fields in `auvsi_suas_accesslog` and stores all of the subtype fields in a subtype table.

When querying for one of the subtypes, Django does a JOIN of the subtype table and `auvsi_suas_accesslog`. The query for `user_on_clock` looks something like:

```sql
SELECT "auvsi_suas_accesslog"."id", "auvsi_suas_accesslog"."timestamp", "auvsi_suas_accesslog"."user_id",
  "auvsi_suas_missionclockevent"."accesslog_ptr_id", "auvsi_suas_missionclockevent"."team_on_clock",
  "auvsi_suas_missionclockevent"."team_on_timeout"
FROM "auvsi_suas_missionclockevent"
INNER JOIN "auvsi_suas_accesslog" ON (
[...]

PR #140, opened 2016-06-23
Add memory for sort operations.

pmtischler:
For #139


prattmic:
LGTM

FWIW, this had no effect on `/api/teams` performance for me. (Assuming I applied it correctly by manually editing `/etc/postgresql/9.3/main/postgresql.conf` and restarting postgres)

pmtischler:
I didn't expect it to have a big effect. The default is 4MB, and we saw ~100MB access log table size. We likely weren't performing a sort on the data- per your other comment it's likely just the join operation. This is more defensive- if we accidentally sort on top of that :)

PR #141, opened 2016-06-26
Convert AccessLog to an abstract base class

prattmic:
Making AccessLog an abstract base class means that rather than having a
dedicated AccessLog table in the database, each subtype's table contains
the AccessLog fields.

In my stress test, this makes the SQL queries for `/api/teams` go from
taking 6600ms total to 120ms.

The change itself is trivial, but the migration is complex, as Django
cannot gracefully handle this kind of migration. See 0018_accesslog.py
for details.

References #139

### DIFF
--- a/server/auvsi_suas/migrations/0019_accesslog_finish.py
+++ b/server/auvsi_suas/migrations/0019_accesslog_finish.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+"""See 0018_accesslog for the first stage of this migration."""
+
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import datetime
+from django.utils import timezone
+from django.conf import settings
+import functools
+import logging
+
+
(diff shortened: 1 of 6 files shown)


server/auvsi_suas/migrations/0019_accesslog_finish.py
```diff
@@ -0,0 +1,71 @@
+# -*- coding: utf-8 -*-
+
+"""See 0018_accesslog for the first stage of this migration."""
```
pmtischler: Could this be merged with previous migration?

pmtischler:
LGTM after comment & travis fixes.

server/auvsi_suas/migrations/0019_accesslog_finish.py
prattmic: No, see the comment in the previous migration

prattmic:
It would have helped if I actually ran the tests...

PTAL

pmtischler:
LG

PR #142, opened 2016-06-26
Enable the profiling panel on the debug toolbar

prattmic:
This gives a nice breakdown of where views are spending their time.
The max depth of 50 ensures we can actually see everything relevant.

References #139


pmtischler:
LGTM after travis fix.

PR #153, opened 2016-07-25
Remove unnecessary upgrade from automated setup.

### LINKED ISSUE #114: Setup.sh upgrades packages without user permission
As the title says, when I run `upgrade.sh` it upgrades packages on my system without my permission by running `apt-get upgrade`. A script should not upgrade my packages without my permission because upgrade packages can, in cases where packages are not held back correctly, break software. 

If a specific version of a package is required, it should be specified in the `apt-get install` as `apt-get install package=version`.

pmtischler:
Fixes #114


prattmic:
LGTM

Issue #144, opened 2016-07-07
Remove non telemetry access logging and rate calculation


pmtischler:
We should remove these elements from the rules to decrease complexity.

prattmic:
:+1:

PR #155, opened 2016-07-25
Remove server info and its access log.

### LINKED ISSUE #149: Remove server info
As an interface and task

pmtischler:
Fixes #149
For #144


prattmic:
LGTM. Docs need to be updated per failing build.

PR #156, opened 2016-07-26
Remove obstacle access log.

### LINKED ISSUE #144: Remove non telemetry access logging and rate calculation

pmtischler:
Fixes #144


prattmic:
LGTM

Issue #143, opened 2016-07-07
Add consistent caching via separate process


pmtischler:
E.g. memcached

This will reduce cross process inconsistencies during invalidation on a change.

prattmic:
Hashing and invalidation of what? The few things we cache in the server?

pmtischler:
Yes, like obstacle configuration used for spline generation.

pmtischler:
Oh I also forgot. We can use the consistent hashing to do things like last position for a team, or whether the team is active.

pmtischler:
Realizing I was for some reason using consistent "hashing" where I meant consistent "caching". My mind must have leaked thoughts from a different project. Sorry about that.

PR #157, opened 2016-07-26
Use Memcached over in-process cache.

pmtischler:
This will create a consistent cache, which removes need to repeatedly
press a cache invalidation button, allows us to do proper invalidation
in the future, and allows us to use the cache for recent event queries
(e.g. latest telemetry data per user).

For #143

### DIFF
--- a/client/interop/client_test.py
+++ b/client/interop/client_test.py
@@ -13,7 +13,9 @@ from . import Client, AsyncClient, InteropError, Target, Telemetry
 # if the defaults are not correct.
 server = os.getenv('TEST_INTEROP_SERVER', 'http://localhost')
 username = os.getenv('TEST_INTEROP_USER', 'testuser')
-password = os.getenv('TEST_INTEROP_PASS', 'testpass')
+password = os.getenv('TEST_INTEROP_USER_PASS', 'testpass')
+admin_username = os.getenv('TEST_INTEROP_ADMIN', 'testadmin')
+admin_password = os.getenv('TEST_INTEROP_ADMIN_PASS', 'testpass')
 
 
 class TestClientLoggedOut(unittest.TestCase):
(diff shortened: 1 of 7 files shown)


setup/puppet_files/auvsi_suas/manifests/memcached_setup.pp
```diff
@@ -0,0 +1,11 @@
+# AUVSI SUAS Puppet Module: memcached_setup
+# Installs and configures memcached for the server.
+# ==============================================================================
+
+class auvsi_suas::memcached_setup {
+    class { 'memcached':
+        package_ensure => 'latest',
+        max_memory => 512,
+        listen_ip => '127.0.0.1',
```
prattmic: Perhaps explicitly set the IP here, since it is hard-coded in settings.py?

prattmic:
From the tests, it looks like the setup is missing a memcache Python module.

setup/puppet_files/auvsi_suas/manifests/memcached_setup.pp
pmtischler: You mean the port? If so, done.

pmtischler:
Done. PTAL

pmtischler:
Odd. It passes on a fresh Vagrant VM.

pmtischler:
Facepalm. I hadn't added the change to the commit. PTAL.

client/interop/client_test.py
```diff
@@ -51,6 +53,11 @@ class TestClient(unittest.TestCase):
 
     def setUp(self):
         """Create a logged in Client."""
+        # Create an admin client to clear cache.
+        client = Client(server, admin_username, admin_password)
+        client.get('/api/clear_cache')
```
prattmic: This seems to be indicative of some sort of cache coherency problem. Why is a manual cache flush necessary?

pmtischler: We do have cache coherency problems with GPS positions. Hopefully our own custom REST API would allow us to properly invalidate in the future.

prattmic: I'm not sure I understand the problem, but can you file a bug and reference it here?

pmtischler: See #162

PR #165, opened 2016-08-21
Remove internal component ids within mission.

pmtischler:
These IDs are internal implementation details not needed by the users of
the interface and provides unnecessary confusion.


prattmic:
LGTM

Issue #148, opened 2016-07-08
Add interface to request mission details


pmtischler:
E.g. fly zones, waypoints, etc

PR #166, opened 2016-08-21
Add mission types to client library.

pmtischler:
For #148

Following this PR I'll send one to extend the client library, and add a simple tool to make a request and print the details.


client/interop/types.py
```diff
@@ -44,7 +44,113 @@ def serialize(self):
...
+
+
+class Waypoint(ClientBaseType):
+    """Waypoint consisting of an order, GPS position, and optional altitude.
+
+    Attributes:
+        order: An ID giving relative order in a set of waypoints.
+        latitude: Latitude in decimal degrees.
+        longitude: Longitude in decimal degrees.
+        altitude: Optional. Altitude in feet MSL.
```
prattmic: When do waypoints omit altitude?

client/interop/types.py
```diff
@@ -44,7 +44,113 @@ def serialize(self):
...
+    Attributes:
+        id: The unique ID of the mission.
+        active: Whether the mission is active.
+        air_drop_pos: The GpsPosition of the air drop.
+        fly_zones: A list of FlyZone boundaries the UAS must be within.
+        home_pos: The GpsPosition of the UAS launch point (tents).
+        mission_waypoints: A list of Waypoint the UAS must traverse.
+        off_axis_target_pos: The GpsPosition of the off-axis target.
+        search_grid_points: List of Waypoint defining the search grid polygon.
+        sric_pos: The GpsPosition of the SRIC.
```
prattmic: Raises: ValueError

prattmic:
LGTM

client/interop/types.py
pmtischler: E.g. for the boundary points:
https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/models/mission_config.py#L361

client/interop/types.py

PR #167, opened 2016-08-26
Add mission request to client.

pmtischler:
For #148

### DIFF
--- a/client/interop/client.py
+++ b/client/interop/client.py
@@ -14,7 +14,10 @@ import requests
 import threading
 
 from .exceptions import InteropError
-from .types import StationaryObstacle, MovingObstacle, Target
+from .types import Mission
+from .types import MovingObstacle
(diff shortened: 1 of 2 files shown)


client/interop/client.py
```diff
@@ -107,6 +110,24 @@ def delete(self, uri):
...
+            requests.Timeout: Request timeout.
+            ValueError or AttributeError: Malformed response from server.
+        """
+        r = self.get('/api/missions')
+        d = r.json()
+
+        missions = []
+        for o in d:
+            missions.append(Mission.deserialize(o))
+        return missions
```
prattmic: Optional:

`return [Mission.deserialize(m) for m in r.json()]`

prattmic:
LGTM

client/interop/client.py
pmtischler: Nice. Done.

PR #168, opened 2016-08-27
Add tool to request mission details and print them.

### LINKED ISSUE #148: Add interface to request mission details

pmtischler:
Fixes #148

### DIFF
--- a/client/tools/README.md
+++ b/client/tools/README.md
@@ -1,18 +1,17 @@
 AUVSI SUAS: Interoperability Client Tools
-================================================================================
+=========================================
 
-## Prober
+get_missions.py
+---------------
 
-The prober is a simple use of the interoperability client which validates the
-system is working and recieving data. Assuming tests passed before deployment,
-this can be used to validate the system is accessible by teams.
+This script can get the mission details from the interoperability server.
 
 ```
-python prober.py [interop_server_host] [interop_time] [username] [password] [generator] [flightsim_kml_path]
-python prober.py http://localhost:80 0.1 testuser testpass zeros
+./get_missions.py --url http://localhost --username testuser
 ```
 
-## Upload Targets
+upload_targets.py
+-----------------
 
 This script can upload targets defined in the old text file "Electronic Target
 Data Format" as defined by the competition rules.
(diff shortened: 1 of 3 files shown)


client/tools/README.md
```diff
@@ -27,5 +26,17 @@ Limitations:
...
+
+prober.py
+---------
+
+The prober is a simple use of the interoperability client which validates the
+system is working and recieving data. Assuming tests passed before deployment,
+this can be used to continuously validate the system is accessible by teams.
+
+```
+python prober.py [interop_server_host] [interop_time] [username] [password] [generator] [flightsim_kml_path]
```
prattmic: prober.py is executable, so just say `./prober.py` like the other examples.

pmtischler: Done.

client/tools/get_missions.py
```diff
@@ -0,0 +1,34 @@
...
+
+from interop import Client
+from interop import Target
+
+
+def main(url, username, password):
+    """Program to get and print the mission details."""
+    client = Client(url, username, password)
+    missions = client.get_missions()
+    print missions
```
prattmic:
```py
from __future__ import print_statement
import pprint

for i, m in enumerate(missions):
    print("Mission %d" % i)
    pprint.pprint(m.serialize())
```

Should make this much easier to read and compatible with Python 3.

prattmic:
One suggestion about making the output easier to read and Python 3 compatibility. Otherwise LGTM

client/tools/get_missions.py
prattmic: Hm, apparently Mission doesn't support serialize. I'd still recommend some kind of custom printing logic. The output must be nigh unreadable as-is, given the amount of stuff in Mission.

pmtischler: I was relying on the string encoding of Mission. I haven't implemented serialization for Mission yet. I've changed it to use print() though.

pmtischler:
Filed #170 to handle the serializing & pretty print. I"m pushing off any work that is non blocking or not needed for team testing until after the rules are released.

client/tools/get_missions.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
+# Program to request and print mission details.
+
+import argparse
+import getpass
+from __future__ import print_statement
```
prattmic: It is print_function, not print_statement, I was dumb. Statement is the old behavior :)

PR #172, opened 2016-09-14
Switch to Docker. Restructure Repo. Remove Puppet.

### LINKED ISSUE #103: Release Docker Containers as Default Deployment
Similar to Vagrant, to decrease time for teams to start using interop.
https://docs.travis-ci.com/user/docker/

pmtischler:
Fixes #103

Add Dockerfiles and utility scripts to create an image for the server
and an image for the client. Use separate images to provide granular
deployment, and allow teams to build custom image from client image.

Restructure repo so subdirectories (server, client, tools) seperately
setup/build and run, mirroring Docker structure.

Remove Puppet because it adds another dependency which has caused
flakiness and breakages. It's less useful with Docker images built from
scratch each time. Docker needs all setup in Dockerfile (not Puppet) in 
order to do incremental builds.

Simplify and update the documentation to match new structure.


pmtischler:
Apologies for the long PR. I'm still updating the docs, but I'd like to submit by Thursday morning so I figured we should start review asap.

pmtischler:
I have upgraded the documentation. PTAL

pmtischler:
@prattmic I'm having trouble getting the Target evaluator test to pass on Travis, despite it working locally on a clean build. Can you take a look?

client/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+        python-pip \
+        python-pyproj \
+        python-virtualenv \
+        python3 \
+        python3-pip \
+        python3-pyproj
+
+RUN mkdir -p /interop/client && \
+    bash -c "cd /interop/client && \
+      virtualenv --system-site-packages -p /usr/bin/python2 venv2 && \
```
prattmic: I think you could omit the `cd` above by just doing `virtualenv --system-site-packages -p /usr/bin/python2 /interop/client/venv2`

client/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# Builds the Interop Client Docker image.
+
+CLIENT=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
+sudo docker build -t auvsisuas/interop.client ${CLIENT}
```
prattmic: Nit: it seems like auvsisuas/interop-client would be a more normal style name

prattmic: I think this should drop `sudo`. It isn't generally required to use docker.

client/configure.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# Configures the Interop Client for the Docker image build.
+
+CLIENT=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
+
+export PYTHONPATH=$CLIENT
+source $CLIENT/venv2/bin/activate
```
prattmic: `${CLIENT}`

client/run.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# Runs the Interop Client in a container.
+
+CLIENT=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
```
prattmic: This is unused

client/run.sh
```diff
@@ -0,0 +1,5 @@
...
+sudo docker run -it auvsisuas/interop.client
```
prattmic: Since this is in a script, it may be clearer to write out the full arg names (`--interactive --tty`)

docs/configuration.rst
```diff
@@ -0,0 +1,41 @@
...
+**Create a New Admin**. The automated system setup will create a default
+administrator account with username ``testadmin`` and password ``testpass``.
+This account can be used to login to the admin interfaces or authenticate any
+requests. If you would like to create another administrator account, the
+Django management program can be used to create an administrator account.
+Execute the following from a bash shell inside the server's container.
+
+.. code-block:: bash
+
+    cd /interop/server
```
prattmic: Don't you still need to enter the venv?

docs/getting_started.rst
```diff
@@ -0,0 +1,510 @@
...
+	  color=blue;
+      label="Judges";
+      subgraph cluster_interop_router {
+        style=filled;
+        color=lightgrey;
+        label="Interop Router: 10.10.130.1";
+        node [style=filled,color=white];
+  	    interop_lan0 [label="LAN 0: 10.10.130.2"];
+	    interop_lan1 [label="LAN 1: 10.10.130.3"];
+		interop_lan1 -- interop_lan0;
```
prattmic: Some of this indentation seems wrong to me (though I've never read graphviz before)

docs/getting_started.rst
```diff
@@ -0,0 +1,510 @@
...
+**Create and Start Container**. The interop server is provided as a Docker
+image and should be run as a Docker container. The repo provides a script to
+run the container in a standard way: it creates the container, runs it in the
+background, uses port ``8000`` for the web server, and restarts automatically
+(e.g. on failure or boot) unless explicitly stopped.
+
+.. code-block:: bash
+
+    cd ~/interop
+    ./server/run.sh
```
prattmic: Does `docker run` download the container if it doesn't already exist?

docs/getting_started.rst
```diff
@@ -0,0 +1,510 @@
...
+management tool), or do other container-local actions, you can start a bash
+shell inside of the container. The following shows how to start the shell.
+
+.. code-block:: bash
+
+    sudo docker exec -it interop.server bash
+
+**Dump Database, Dump Server Log**. The shell will start the user inside of the
+working directory (server source code) at ``/interop/server``. The following
+shows how to dump the database to standard output, dump the server log file to
```
prattmic: Why would you want a database dump on stdout?

docs/getting_started.rst
```diff
@@ -0,0 +1,510 @@
...
+shell. The following shows how to execute it for the default testing user
+(``testuser``) if the interop server was at ``192.168.1.2:8000``.
+
+.. code-block:: bash
+
+    cd /interop/client
+    ./tools/get_missions.py --url http://192.168.1.2:8000 \
+        --username testuser --password testpass
+
+**Upload Targets**. The client image provides a script to upload targets to the
```
prattmic: I think all of these headings would be better as subsection heading, because then you could link directly to them.

docs/getting_started.rst
```diff
@@ -0,0 +1,510 @@
...
+browser. You may need to replace ``localhost:8000`` if you've configured the
+setup differently. This will prompt for an admin account login, so enter the
+preconfigured user: ``testadmin`` with password ``testpass``.
+
+**SUAS Admin Dashboard**. After login it will show the SUAS made admin
+dashboard. It will have a navigation bar with system-wide and mission-specific
+links. The homagepage for the dashboard will also list the current missions,
+and should show the single mission which comes with the image. If you click the
+mission, you will be brought to a mission-specific dashboard. Click the "Help"
+button on the mission dashboard to learn how to use this interface.
```
prattmic: I know this says 'admin', but it is probably worth re-iterating that this is for development and testing. Teams will not have access to this during the mission.

There was some confusion about that last year.

docs/getting_started.rst
```diff
@@ -0,0 +1,510 @@
...
+
+The following is an example of how to perform interoperability using the
+``curl`` command. **This is too inefficient to achieve a sufficient
+update rate**. This merely shows how simple it is to implement
+interoperability from standard HTTP and JSON.
+
+The curl command has the following parameters:
+
+#. **-b**: The place to save received cookies.
+#. **-c**: The place to load and send cookies.
```
prattmic: These descriptions don't seem quite right based on the man page.

Also, consider using `--cookie` and `--cookie-jar`

docs/getting_started.rst
```diff
@@ -0,0 +1,510 @@
...
+
+    target = interop.Target(type='standard',
+                            latitude=38.145215,
+                            longitude=-76.427942,
+                            orientation='n',
+                            shape='square',
+                            background_color='green',
+                            alphanumeric='A',
+                            alphanumeric_color='white')
+    target = client.post_target()
```
prattmic: `client.post_target(target)`

docs/integration/client.rst
```diff
@@ -1,16 +0,0 @@
-Interoperability Client
```
prattmic: Why drop the client docs?

docs/specification.rst
```diff
@@ -1,5 +1,5 @@
-Interoperability Specification
```
prattmic: The file is not in this commit, but perhaps update the version in conf.py to 2017?

server/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+RUN sudo ln -snf /usr/bin/nodejs /usr/local/bin/node && \
+    sudo npm install -g \
+        karma \
+        karma-chrome-launcher \
+        karma-jasmine \
+        karma-phantomjs-launcher \
+        phantomjs
+
+RUN pip install \
```
prattmic: Use the requirements file.

prattmic: No more venv? I guess it isn't really necessary since we're in a dedicated container.

server/configure.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# Configures the Interop Server for the Docker image build.
+
+SERVER=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
+
+set -e
```
prattmic: Will a non-zero exit code of this script when run via RUN cause the docker build to fail? I'd hope so, but the Docker docs aren't clear.

tools/setup_docker.sh
```diff
@@ -0,0 +1,26 @@
...
+sudo apt-get -qq update
+
+log "Installing APT packages for Docker."
+sudo apt-get -qq install -y \
+    docker-engine \
+    linux-image-extra-$(uname -r) \
+    linux-image-extra-virtual
+
+log "Starting Docker daemon."
+sudo service docker restart
```
prattmic: Usually installing a service automatically starts it. Is that not the case with docker?

client/build.sh
zyeri: Not all Linux distributions have their Docker packages automatically add the user to the `docker` group (allowing for `sudo`-less usage) and those that do are creating a major security hole in the system as the `docker` group is root equivalent. See docker/docker#9976 and [Docker Security](https://docs.docker.com/engine/security/security/).

prattmic: Sure, I'm not necessarily advocating for having everyone allow sudo-less docker, but this script can be executed with sudo, rather than containing sudo.

client/requirements.txt
```diff
@@ -1,3 +0,0 @@
-futures
```
pmtischler: The problem with this here and below is that Docker won't have access to the requirements file unless you COPY it to the image. Once you do that, any change to the folder will invalidate anything after the COPY statement, which will include all installation of dependencies.

client/Dockerfile
pmtischler: Done.

client/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+      source /interop/client/venv2/bin/activate && \
+      pip2 install \
```
pmtischler: See previous comment.

client/requirements.txt
prattmic: The Docker best practices doc actually includes exactly this example, where they initially copy *only* the requirements file, then copy everything else later.

https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/add-or-copy

pmtischler: Done.

client/Dockerfile
pmtischler: Done.

client/build.sh
pmtischler: Done.

pmtischler: Removed sudo, added sudo to docs+travis for execution call.

client/configure.sh
pmtischler: Done.

client/run.sh
pmtischler: Done.

docs/configuration.rst
pmtischler: Now that we are in an image I removed venv for the server.

docs/getting_started.rst
pmtischler: Somehow tabs snuck in. Fixed it

docs/getting_started.rst
pmtischler: Yes

docs/getting_started.rst
pmtischler: More as a usage example for `manage.py`, rather than something you actually want to do at this point.

docs/getting_started.rst
pmtischler: Added to preconfigured users paragraph.

docs/getting_started.rst
pmtischler: Done.

docs/getting_started.rst
pmtischler: Done.

docs/integration/client.rst
pmtischler: I've flattened the structure, and created a page specifically for the interop client libs.

docs/specification.rst
pmtischler: Done.

server/Dockerfile
pmtischler: Yup

pmtischler: Done.

server/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+        pyproj \
+        python-memcached \
+        scipy \
+        "simplekml==1.2.7"
+
+COPY . .
+
+RUN /interop/server/configure.sh
+
+VOLUME /var/log/apache2 /var/lib/postgresql/9.3/data
```
pmtischler: Afaik, this makes it persist across start/stop. It's a list of directories in the container, rather than a mapping of host to container. The way to do host-container mapping is at ``docker run`` command line flags.

server/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# Builds the Interop Server Docker image.
+
+SERVER=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
+sudo docker build -t auvsisuas/interop.server ${SERVER}
```
pmtischler: Done.

server/configure.sh
pmtischler: Yes, I believe so.

tools/setup_docker.sh
pmtischler: Doesn't seem so, this was recommended in the install docs. I used restart so the script could be executed multiple times with a clean exit.

server/auvsi_suas/models/target_test.py
```diff
@@ -331,7 +332,7 @@ def setUp(self):
                               location=l1,
                               orientation=Orientation.n,
                               shape=Shape.circle,
-                              background_color=Color.white,
```
pmtischler: It was failing before this change. I made this change to make it even more obvious the two should be paired. I've reverted this change.

docs/getting_started.rst
pmtischler: I did it this way because eng writing style (or at least one I've been taught) prefers to have at least 2 paragraphs for a section and at least 2 subsections, as otherwise it's unnecessary structure that provides more noise at the TOC level than signal.

Linking can be achieved without creating subsections. Though I've never linked to these docs beyond the page level, so I'm not sure it's worth the effort to add.

I don't feel strongly, I can make that change if you prefer it.

prattmic: Well, this page is huge, hence why I thought it could be useful to provide direct links, regardless of how.

I don't feel too strongly, you'll probably be directing people to parts of the docs more than me.

.travis.yml
```diff
@@ -2,16 +2,32 @@ sudo: required
...
-  # Travis CI does not make the home directory world readable by default,
-  # which prevents Apache from reading the server source.
+  # Make home dir world readable so Apache can read server source.
   - sudo chmod 755 /home/travis/
-  - sudo ./setup/setup.sh
+  # Setup tools.
+  - sudo ./tools/setup_tools.sh
+  # Build the images.
+  - sudo  ./server/build.sh
+  - sudo ./client/build.sh
```
prattmic: AFAICT from the Travis docs, docker does not need to be run as a superuser.

https://docs.travis-ci.com/user/docker/

docs/getting_started.rst
```diff
@@ -0,0 +1,513 @@
...
+
+The following is an example of how to perform interoperability using the
+``curl`` command. **This is too inefficient to achieve a sufficient
+update rate**. This merely shows how simple it is to implement
+interoperability from standard HTTP and JSON.
+
+The curl command has the following parameters:
+
+#. **--cookie**: The place to save received cookies.
+#. **--cookie-jar**: The place to load and send cookies.
```
prattmic:
```
#. **--cookie**: Cookies in this file are sent to the server.
#. **--cookie-jar**: Cookies sent from the server are saved in this file.
```

docs/integration/client.rst
prattmic: Oops, I missed that before writing this comment.

server/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+        npm \
+        postgresql \
+        postgresql-client \
+        python \
+        python-matplotlib \
+        python-numpy \
+        python-pip \
+        python-psycopg2 \
+        python-pyproj \
+        python-scipy
```
prattmic: Comment to note that we install some of these from apt just because it is faster?

server/README.md
```diff
@@ -11,9 +11,7 @@ Django project, with the following components:
   3. `manage.py`: A utility program provided by Django which can be used to
      manage the application. This includes creating databases, running the
      development server, etc.
-  4. `requirements.txt`: The dependencies which are installed as part of
```
prattmic: Revert

docs/conf.py
```diff
@@ -109,12 +110,23 @@
...
-autodoc_mock_imports = ['concurrent.futures', 'dateutil.parser', 'requests']
+autodoc_mock_imports = [
+    'concurrent.futures',
+    'dateutil.parser',
+    'django',
+    'django.conf',
+    'networkx',
+    'numpy',
+    'pyproj',
+    'pyproj.Proj',
```
prattmic: I think this is breaking the doc build

client/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+COPY requirements.txt requirements.txt
+RUN bash -c "virtualenv --system-site-packages -p /usr/bin/python2 \
+            /interop/client/venv2 && \
+        source /interop/client/venv2/bin/activate && \
+        pip install -r requirements.txt" && \
+    bash -c "virtualenv --system-site-packages -p /usr/bin/python3 \
+            /interop/client/venv3 && \
+        source /interop/client/venv3/bin/activate && \
+        pip3 install -r requirements.txt"
```
prattmic: You either need nosetests here or in requirements.txt for testing.

docs/getting_started.rst
pmtischler: Done.

server/Dockerfile
pmtischler: Done.

server/README.md
pmtischler: Done, but simplified it.

docs/conf.py
pmtischler: The problem is that pyproj isn't in the tool venv. I think this'll be cleaner to just have it imported. I've updated the tools to require the venvs from the server and client.

client/Dockerfile
pmtischler: Isn't that covered by nose? The tests run in the Docker client image, so I assume it's covered.

pmtischler: Spoke too soon, for some reason it's failing now.

pmtischler: Done.

pmtischler:
This PR has been fixed. @prattmic PTAL

server/server/settings.py
```diff
@@ -193,7 +194,7 @@
 TEST_ENABLE_LOADTEST = True
 
 # The time to execute each loadtest for
-TEST_LOADTEST_TIME = 10.0
+TEST_LOADTEST_TIME = 2.0
```
prattmic: Did you mean to change this?

tools/requirements.txt
```diff
@@ -0,0 +1,7 @@
+-r ../client/requirements.txt
+-r ../server/requirements.txt
```
prattmic: Bleh, I'm not a big fan of this. Not a huge deal though.

server/server/settings.py
pmtischler: Yes, the 30 wait was a little excessive.

PR #180, opened 2016-10-04
Make mission serializable. Use pretty print in get_missions.

### LINKED ISSUE #170: Make Mission Serializable  and Pretty Print
Discussion in #168 E.g. 
```
import pprint
pprint.pprint(m.serialize())
```

pmtischler:
Fixes #170

### DIFF
--- a/client/interop/types.py
+++ b/client/interop/types.py
@@ -38,8 +38,16 @@ class ClientBaseType(object):
 
     def serialize(self):
         """Serialize the current state of the object."""
-        return {k: self.__dict__[k]
-                for k in self.attrs if self.__dict__[k] is not None}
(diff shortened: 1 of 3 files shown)


client/interop/types.py
```diff
@@ -38,8 +38,17 @@ def __unicode__(self):
 
     def serialize(self):
         """Serialize the current state of the object."""
-        return {k: self.__dict__[k]
-                for k in self.attrs if self.__dict__[k] is not None}
+        serial = {}
+        for attr in self.attrs:
+            data = self.__dict__[attr]
+            if data is not None:
```
prattmic: Why not include None objects in the serialized dict?

Issue #176, opened 2016-09-16
Update target upload


pmtischler:
Use the new Object File Format

pmtischler:
Consider handling during #171

pmtischler:
@prattmic Can you update the [docs](http://auvsi-suas-competition-interoperability-system.readthedocs.io/en/latest/getting_started.html#auvsisuas-interop-client) as well?

PR #191, opened 2016-10-13
Update upload tool

prattmic:
There are two commits in this PR. The first adds support for uploading via the 2017 object file format. The second adds a helpful mode that fetches the uploaded targets.

### DIFF
--- a/client/tools/interop_cli.py
+++ b/client/tools/interop_cli.py
@@ -13,7 +13,7 @@ import time
 from interop import Client
 from interop import Target
 from interop import Telemetry
-from upload_targets import upload_targets
+from upload_targets import upload_targets, upload_legacy_targets
 
 
 def missions(args, client):
@@ -23,7 +23,10 @@ def missions(args, client):
 
 
 def targets(args, client):
-    upload_targets(client, args.target_filepath, args.imagery_dir)
+    if args.legacy_filepath:
+        upload_legacy_targets(client, args.legacy_filepath, args.target_dir)
+    else:
+        upload_targets(client, args.target_dir)
 
 
 def probe(args, client):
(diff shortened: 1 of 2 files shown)


client/tools/interop_cli.py
```diff
@@ -23,7 +25,16 @@ def missions(args, client):
 
 
 def targets(args, client):
-    upload_targets(client, args.target_filepath, args.imagery_dir)
+    if args.legacy_filepath:
+        if not args.target_dir:
+            raise ValueError('--target_dir is required.')
+        upload_legacy_targets(client, args.legacy_filepath, args.target_dir)
```
pmtischler: Do we want to provide legacy support? Seems like a lot of effort on our end, and given how early it is in the year they have more than enough time to adapt.

client/tools/upload_targets.py
```diff
@@ -129,3 +135,66 @@ def upload_targets(client, target_filepath, imagery_dir):
...
+    for entry in os.listdir(target_dir):
+        name, ext = os.path.splitext(entry)
+
+        if ext.lower() == '.json':
+            if name in targets:
+                raise ValueError(
+                    'Found duplicate target files for %s: %s and %s' %
+                    (name, targets[name], entry))
+            targets[name] = os.path.join(target_dir, entry)
+        elif ext.lower() == '.png' or ext.lower() == '.jpg' or ext.lower(
```
pmtischler: Can you split lines on the OR instead?

pmtischler:
LGTM

client/tools/interop_cli.py
prattmic: I left the legacy support specifically because it was only a few lines to support. The code is already there. There is little reason to remove it at this point.

client/tools/upload_targets.py
prattmic: Done

client/tools/upload_targets.py
```diff
@@ -129,3 +135,66 @@ def upload_targets(client, target_filepath, imagery_dir):
...
+    """
+    targets = {}
+    images = {}
+
+    for entry in os.listdir(target_dir):
+        name, ext = os.path.splitext(entry)
+
+        if ext.lower() == '.json':
+            if name in targets:
+                raise ValueError(
```
prattmic: Many filesystems are case-sensitive. There may be 1.json and 1.JSON.

pmtischler:
For #176

PR #194, opened 2016-10-15
Cleanup spec sample JSON

prattmic:
Trailing commas are not allowed in JSON, and emergent_last_known_pos was
missing from GET /api/missions.


prattmic:
@pmtischler The sample file in the rules appendix also has a trailing comma that it should not.

pmtischler:
LGTM

pmtischler:
LGTM

Issue #145, opened 2016-07-07
Add per team evaluation output


pmtischler:
This will allow for incremental scoring at competition rather than daily batch.

PR #196, opened 2016-10-17
Breakup backend js and inline into pages.

pmtischler:
For #145.

This is necessary to create interfaces which don't want the underlying data updating on an interval (e.g. select a team to review, which will erase choices after an update). This makes the target interface more intuitive (it doesn't update beneath you).


server/auvsi_suas/static/auvsi_suas/pages/mission-list/mission-list-controller.js
```diff
@@ -14,9 +13,13 @@
  */
 MissionListCtrl = function(Backend) {
     /**
-     * @export @const {!Object} The backend service.
+     * @private {?Array<Object>} The missions data.
```
prattmic: Out of curiosity, does something actually check these types?

PR #198, opened 2016-10-22
Return telemetry in teams request. Index (user,timestamp) for telemetry.

### LINKED ISSUE #164: Make latest telemetry queries more efficient
We hit a scalability problem with latest queries for telemetry. @prattmic improved the db query substantially, which will help us survive. We can use the now consistent cache to store the result of "latest" queries, like latest telemetry for each user.

pmtischler:
Returning telemetry in the teams request makes fewer requests and
database queries.

Indexing (user,timestamp) should allow simultaneous user filtering,
timestamp filtering, and ordering by timestamp via the index without
having to post-filter and post-sort. This should make requests more
efficient.

Fixes #164

### DIFF
--- a/server/auvsi_suas/models/access_log_test.py
+++ b/server/auvsi_suas/models/access_log_test.py
@@ -50,7 +50,7 @@ class TestAccessLogCommon(TestCase):
             uas_position.save()
             log = UasTelemetry(user=user,
                                uas_position=uas_position,
-                               uas_heading=0)
+                               uas_heading=0.0)
             log.save()
             log.timestamp = start + i * delta
             log.save()
@@ -76,7 +76,7 @@ class TestAccessLogBasic(TestAccessLogCommon):
         self.assertEqual(len(logs), 0)
 
         logs = UasTelemetry.by_time_period(self.user1, [])
-        self.assertEqual(len(logs), 0)
+        self.assertSequenceEqual([], logs)
 
         log_rates = UasTelemetry.rates(self.user1, [])
         self.assertTupleEqual(log_rates, (None, None))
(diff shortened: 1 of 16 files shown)


server/auvsi_suas/models/access_log_test.py
```diff
@@ -104,36 +104,39 @@ def test_multi_user(self):
...
+                                       delta=delta)
+
+        logs = UasTelemetry.by_user(self.user1,
+                                    start_time=start,
+                                    end_time=start + delta * 10)
+        self.assertSequenceEqual(expect_logs, logs)
+
+        logs = UasTelemetry.by_user(self.user1, start_time=start + delta * 11)
+        self.assertSequenceEqual([], logs)
+        logs = UasTelemetry.by_user(self.user1, end_time=start - delta)
```
prattmic: You could get the same results with end_time=start

server/auvsi_suas/views/teams_test.py
```diff
@@ -352,12 +366,12 @@ def test_id_ignored(self):
 
         self.assertEqual(expected, data['id'])
 
-    def test_active_ignored(self):
-        """active field ignored"""
+    def test_telemetry_ignored(self):
+        """telemetryfield ignored"""
```
prattmic: telemetry field

pmtischler: Done.

server/auvsi_suas/models/access_log_test.py
pmtischler: Done.

pmtischler:
I'm attempting to evaluate now. I pulled the data off of the interop laptop and uploaded to our Drive folder. Import went quick, but migration is taking a long time (still on the first, migration 18). Given how long this is taking, I'll upload the migrated form as well to save us time in future testing.

pmtischler:
Test completed. Was able to look at admin interface and CPU stayed under 15% peak, with about ~10% average.

PR #199, opened 2016-10-24
Add admin endpoint to edit mission waypoints

JamesSto:
CUAir always spends a lot of time at test flights copying over our mission waypoints into the interop server when we want to use it, so I added an admin endpoint that accepts a JSON to edit the mission waypoints of a particular mission. This will save valuable time at test flights and hopefully other teams will find being able to automate transferring waypoints from a ground station to the interop server useful as well.

I added a new class to the missions.py file with a POST superuser endpoint that accepts a JSON in the form {"pk":PK, "waypoints":[]}. This creates new GpsPosition, AerialPosition, and Waypoint objects then changes the list of waypoints in the selected mission pk to be the waypoints sent.

I'm happy to add documentation for this change as well in the docs.

### DIFF
--- a/server/auvsi_suas/views/missions.py
+++ b/server/auvsi_suas/views/missions.py
@@ -12,6 +12,7 @@ from django.http import HttpResponse
 from django.http import HttpResponseBadRequest
 from django.http import HttpResponseNotFound
 from django.http import HttpResponseServerError
+from django.http import JsonResponse
 from django.utils.decorators import method_decorator
 from django.views.generic import View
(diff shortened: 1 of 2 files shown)


pmtischler:
This isn't the direction we want to go. If we are going to add a JSON endpoint it's going to be a full mission endpoint, as part of the effort to add the UI components.

I also told you off Github we weren't going to do it this way, and that the best way for you to automate was to use a server local script which directly modified the database.

PR #201, opened 2016-10-26
Add doc on local Python access for automation.

### LINKED ISSUE #200: Write guide explaining programmatic access
For data creation and analysis. This will largely be a tutorial to build something like [create_teams](https://github.com/auvsi-suas/interop/blob/master/server/tools/team_creator/create_teams.py).

@JamesSto This will show you how to automate creating waypoints to speed up testing using existing functionality.

This comes from the offline discussion and subsequent PR. My concern is that nice-to-have code which is added is code we need to maintain, which may represent a long-term burden to the project and hinder time to critical adaptations (e.g. open issues in current milestone). As a pure-volunteer project, we also need to be careful about how much scope we take on.

To be clear- if there is a sufficiently alternative that exists today, it isn't worth the burden to us maintainers.

If web-based JSON access truly is much better for your team than local Python access, forking the repository and adding your team-specific features is always an option. The difference is that you become responsible for maintaining the nice-to-haves your team wants, and the local Python access requires no maintenance.

pmtischler:
Fixes #200.

@JamesSto FYI

### DIFF
--- a/docs/automation.rst
+++ b/docs/automation.rst
@@ -1,0 +1,54 @@
+Automation
+==========
+
+This section describes how to write admin automation for the interop server by
+writing scripts which connect directly to the database and bypass the
+webserver. Note that the database and Django configurations only permit local
+access, so you'll need to run any such scripts locally. This may be useful
+while testing to automatically setup test cases (e.g. test mission) or analyze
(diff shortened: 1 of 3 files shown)


prattmic:
LGTM

PR #202, opened 2016-10-27
Add waypoint automation script

JamesSto:
Modified the endpoint code slightly to be a local automation script instead. Changes waypoints in active mission or given mission PK to match the input JSON.

### DIFF
--- a/server/tools/waypoint_editor/edit_mission_waypoints.py
+++ b/server/tools/waypoint_editor/edit_mission_waypoints.py
@@ -1,7 +1,12 @@
+from __future__ import print_function
+
 import os
 import sys
-sys.path.append("../..")
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "server.settings")
+this_dir = os.path.dirname(os.path.abspath(__file__))
+server_dir = os.path.join(this_dir, '..', '..')
+sys.path = [server_dir] + sys.path
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'server.settings')
 from django.core.wsgi import get_wsgi_application
 get_wsgi_application()
 
@@ -11,27 +16,38 @@ import auvsi_suas.models as models
 
 
 def active_mission():
-    '''Returns the current active mission - modified from views/mission.py'''
+    '''Get the active mission.
+
+    Returns:
+        MissionConfig for active mission or None if there is no activate mission.
+
+    Raises:
+        ValueError: The server has multiple active missions.
+'''
     missions = models.mission_config.MissionConfig.objects.filter(
         is_active=True)
-    if len(missions) != 1:
-        print('Error: Invalid number of active missions. Missions: %s.',
-              str(missions))
-        return None
+    if len(missions) > 1:
+        raise ValueError('Multiple active missions: %s' % missions)
 
-    return missions[0]
+    return missions[0] if missions else None
 
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(
-        usage=
-        'Edits an interop server mission to include the waypoints specified in a JSON. See the example JSON for reference\nUse the --help flag for more information')
-    parser.add_argument("json_file",
-                        help='The file to load the mission waypoints from')
+        description=
+        '''Edits an interop server mission to include the waypoints specified in a JSON.
+
+See the example_waypoints.json for reference.''',
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+
+    parser.add_argument('json_file',
+                        help='The file to load the mission waypoints from',
+                        nargs='?',
+                        default='/dev/stdin')
     parser.add_argument(
         '--pk',
         type=int,
-        default=0,
+        default=None,
         help='Edit the specified pk rather than the currently active mission')
 
     args = parser.parse_args()
@@ -42,7 +58,7 @@ if __name__ == '__main__':
             wp_string = f.read()
     except IOError as e:
         print('Error: Could not read JSON file')
-        print e
+        print(e)
         sys.exit(1)
 
     # Convert json string to list of waypoints
@@ -52,26 +68,13 @@ if __name__ == '__main__':
         print('Error: File is not a valid json')
         sys.exit(1)
 
-    # Confirm all arguments are formatted correctly
-    try:
-        assert type(args.pk) == int
-        assert type(waypoints) == list
-        assert all(type(w) == dict and 'latitude' in w and 'longitude' in w and
-                   'altitude_msl' in w for w in waypoints)
-    except (KeyError, AssertionError) as e:
-        print(
-            'Error: JSON does not contain properly formatted fields or invalid pk format'
-        )
-        print e
-        sys.exit(1)
-
     # Get the mission object to edit
-
     # If the pk is 0, edit the currently active mission
-    if args.pk == 0:
+    if args.pk == None:
         mission = active_mission()
         if mission == None:
-            print 'Error: Active mission not found or multiple active missions'
+            print(
+                'Error: Active mission not found or multiple active missions')
             sys.exit(1)
     else:
         try:
@@ -100,3 +103,4 @@ if __name__ == '__main__':
     # Set the mission waypoints to the new list of objects and save it
     mission.mission_waypoints = waypoints_objects
     mission.save()
+    print('Successfully saved mission waypoints')
--- a/server/tools/waypoint_editor/edit_mission_waypoints.py
+++ b/server/tools/waypoint_editor/edit_mission_waypoints.py
@@ -1,0 +1,102 @@
+import os
+import sys
+sys.path.append("../..")
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "server.settings")
+from django.core.wsgi import get_wsgi_application
+get_wsgi_application()
+
+import argparse
+import json
+import auvsi_suas.models as models
+
+
+def active_mission():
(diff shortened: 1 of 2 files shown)


server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
+import os
+import sys
+sys.path.append("../..")
```
prattmic: Do this like create_teams (https://github.com/auvsi-suas/interop/blob/master/server/tools/team_creator/create_teams.py#L12), using the absolute path of the directory this file is in so that this script will work when run from anywhere.

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+
+    return missions[0]
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        usage=
+        'Edits an interop server mission to include the waypoints specified in a JSON. See the example JSON for reference\nUse the --help flag for more information')
+    parser.add_argument("json_file",
+                        help='The file to load the mission waypoints from')
```
prattmic: `nargs=?, default="/dev/stdin"` to make this optional and default to stdin when omitted.

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+    parser.add_argument(
+        '--pk',
+        type=int,
+        default=0,
```
prattmic: `default=None`

0 is a perfectly valid primary key.

Note that the type conversion won't be applied to `None`: https://docs.python.org/3/library/argparse.html#default

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+        print(
+            'Error: JSON does not contain properly formatted fields or invalid pk format'
+        )
+        print e
+        sys.exit(1)
+
+    # Get the mission object to edit
+
+    # If the pk is 0, edit the currently active mission
+    if args.pk == 0:
```
prattmic: `if args.pk is None`

with the comment above to change default to `None`.

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+    # Convert json string to list of waypoints
+    try:
+        waypoints = json.loads(wp_string)
+    except ValueError:
+        print('Error: File is not a valid json')
+        sys.exit(1)
+
+    # Confirm all arguments are formatted correctly
+    try:
+        assert type(args.pk) == int
```
prattmic: This is unnecessary, argparse will convert this to an int, since you passed type to `add_argument`.

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+        assert type(waypoints) == list
```
prattmic: Please don't do type assertions. The for loop below is sufficient to ensure that waypoints is iterable.

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+        assert all(type(w) == dict and 'latitude' in w and 'longitude' in w and
```
prattmic: Please no type assertions.

I think this would be more readable as a normal loop:

```py
for w in waypoints:
    assert 'latitude' in w
    assert 'longitude' in w
    assert 'altitude_msl' in w
```

Even with this, there are still no checks that the values are numbers. There are two options:

1. Perform more complete input validation. e.g., see https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/views/targets.py#L48-L112

2. Skip all of this validation and allow the code below to raise an exception when something is wrong. Given that this is a simple script, I'd be fine with that.

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
+import os
+import sys
+sys.path.append("../..")
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "server.settings")
+from django.core.wsgi import get_wsgi_application
+get_wsgi_application()
+
+import argparse
```
prattmic: `from __future__ import print_function`

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+              str(missions))
+        return None
+
+    return missions[0]
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        usage=
+        'Edits an interop server mission to include the waypoints specified in a JSON. See the example JSON for reference\nUse the --help flag for more information')
```
prattmic:
```py
parser = argparse.ArgumentParser(
    description='''Edits an interop server mission to include the waypoints specified in a JSON.

See the example_waypoints.json for reference.''',
    formatter_class=argparse.RawDescriptionHelpFormatter)
```

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+    parser.add_argument("json_file",
```
prattmic: Please be consistent about using either " or ' in the file.

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+from django.core.wsgi import get_wsgi_application
+get_wsgi_application()
+
+import argparse
+import json
+import auvsi_suas.models as models
+
+
+def active_mission():
+    '''Returns the current active mission - modified from views/mission.py'''
```
prattmic:
```py
'''Get the active mission.

Returns:
    MissionConfig for active mission or None if there is no activate mission.

Raises:
    ValueError: The server has multiple active missions.
'''
```

server/tools/waypoint_editor/edit_mission_waypoints.py
```diff
@@ -0,0 +1,102 @@
...
+    missions = models.mission_config.MissionConfig.objects.filter(
+        is_active=True)
+    if len(missions) != 1:
```
prattmic: Multiple activate missions is an invalid server configuration. I'd raise an exception in that case:

```py
if len(missions) > 1:
    raise ValueError('Multiple active missions: %s' % missions)

return missions[0] if missions else None
```

JamesSto:
Updated to include changes requested by @prattmic

pmtischler:
As I described in the previous PR and in issue #200, I don't think we want to maintain this. I don't see it providing significant benefit beyond the combination of web interface, programmatic interface, and fixtures provided by Django (for which we maintain nothing). Afaik nothing has changed since the previous decision.

@JamesSto What problem are you trying to solve?

JamesSto:
So our problem is that at test flights, we create a mission in our ground station then want to transfer that mission to the interop server for testing. Entering a new mission, or even just waypoints, can take us up to 20 minutes or so because the web interface just isn't particularly user friendly and entering specific GPS coordinates takes a long time. 

This takes away a lot of valuable time at test flights that could be flight time, and so often it gets pushed to the side, and was a part of the reason why at least for CUAir the interop server connection code didn't undergo much rigorous testing until later than it should have. 

This script is an attempt to reduce the amount of time that would be spent at test flights - we've added functionality to our ground station to export the current mission as a JSON which can be loaded into this program (back when this was in the form of an admin endpoint it could simply POST the JSON to the server rather than needed to export the JSON file then move it to the computer running interop to be put into this script, which was why I found the endpoint preferable).

We thought this functionality could be useful to other teams as well, but if you don't think it's something that's worth maintaining or would be useful to other teams, that's fine - we can just use this locally.

pmtischler:
I don't disagree that this is useful. I do think it can help save you time, and potentially that of other teams. My point is more about the cost-benefit ratio of the script, and that it might be bloat.

I agree that the 3 levels to specify a waypoint is a bit annoying. There probably isn't much harm in just flattening to a single type and not using certain fields in certain contexts. This would have happened as part of PostGIS/GeoDjango migration, which we've postponed. @prattmic Thoughts?

I don't see why it would take up valuable test-flight time- you can set it up ahead of the flight day and you don't need the waypoints set on a MissionConfig to collect data at a test flight (only need before evaluation). I'm also not sure a 20min data-entry process is sufficiently bad to be root-cause of insufficient testing :)

As an alternative, could you request in the reverse direction? We provide [/api/missions](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/views/urls.py#L30). You could have a script which requests the mission details from interop and sets your ground station based on it, rather than the reverse (set interop based on your ground station).

As an alternative, have you tried [fixtures](https://docs.djangoproject.com/en/1.10/howto/initial-data/)? You can specify the raw object as JSON, and then Django provides a command (`django loaddata fixture.json`) to load it. We use this to prime the database with [initial data](https://github.com/auvsi-suas/interop/blob/master/server/fixtures/test_fixture.yaml), and to load [test data](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/fixtures/testdata/sample_mission.json) for e2e testing. You can have your ground station export this JSON and use a script to handle the send & load.

Issue #175, opened 2016-09-16
Update target scoring


pmtischler:
There is a new breakdown of percentages. 
Provide means for judges to upload on behalf of team, give more points for interop submission.

pmtischler:
Reminder that we need to add a actionable flag of some sort to admin use of the interop_cli upload.

PR #204, opened 2016-11-01
Update target scoring

yogishadixit:
* Update match_value function to take into account new scoring scheme
* Update location scoring to be linear
* Update classification scoring 
* Update actionable scoring such that target is only actionable if it is submitted during first flight. Remove classification/location requirements for actionable targets
* Add `interop_submission` field to target to keep track of whether the target was submitted via the interop system. A target is marked as being submitted via the interop system if it is uploaded/updated while the team is still on mission clock. (Any targets uploaded/updated after then are probably a judge uploading from USB submission)
* Change weighting of location, classification and actionable scores. Take into account interop score and autonomy score when calculating match value
* Remove the max 6 autonomous targets restriction
* Update tests

**Note:** There are two rule changes that are not reflected in this PR: (1) counting only the higher-scoring object if both the autonomously and manually submitted objects match the same real target and (2) the extra object penalty. Looking at the output from evaluate_teams, it seems as though these changes should be implemented in the auto-scoring system that we will eventually write. 

**Still TODO:** We need to provide an admin flag to the interop_cli to mark whether the upload is actionable, and accept such in the view code for admins only. We need this to upload USB submissions that were actionable.

### DIFF
--- a/server/auvsi_suas/models/mission_config_test.py
+++ b/server/auvsi_suas/models/mission_config_test.py
@@ -251,17 +251,25 @@ class TestMissionConfigModelSampleMission(TestCase):
         self.assertAlmostEqual(0.5, teams[user0]['uas_telem_times']['max'])
         self.assertAlmostEqual(1. / 6, teams[user0]['uas_telem_times']['avg'])
 
-        self.assertEqual(
-            11, teams[user0]['targets']['manual']['matched_target_value'])
+        self.assertAlmostEqual(
+            0.28,
+            teams[user0]['targets']['manual']['matched_target_value'],
+            places=3)
         self.assertEqual(
             1, teams[user0]['targets']['manual']['unmatched_target_count'])
+
+        self.assertAlmostEqual(
+            1,
+            teams[user0]['targets']['auto']['matched_target_value'],
+            places=3)
+        self.assertEqual(
+            0, teams[user0]['targets']['auto']['unmatched_target_count'])
+
         # Real targets are PKs 1, 2, and 3.
         self.assertEqual(
-            'objective',
-            teams[user0]['targets']['manual']['targets'][1]['actionable'])
+            0.0, teams[user0]['targets']['manual']['targets'][1]['actionable'])
         self.assertEqual(
-            None,
-            teams[user0]['targets']['manual']['targets'][2]['actionable'])
+            1.0, teams[user0]['targets']['auto']['targets'][1]['actionable'])
 
         self.assertEqual(True, teams[user0]['stationary_obst_collision'][25])
         self.assertEqual(False, teams[user0]['stationary_obst_collision'][26])
--- a/server/auvsi_suas/models/target.py
+++ b/server/auvsi_suas/models/target.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from django.db import models
 from gps_position import GpsPosition
 from takeoff_or_landing_event import TakeoffOrLandingEvent
+from mission_clock_event import MissionClockEvent
 
 
 class Choices(enum.IntEnum):
@@ -260,9 +261,8 @@ class Target(models.Model):
     def actionable_submission(self, flights=None):
         """Checks if Target meets Actionable Intelligence submission criteria.
 
-        A target is "actionable" if the aircraft was in flight from initial
-        target submission until the last update. Note that this does not check
-        the localization or characterization Actionable Intelligence criteria.
+        A target is "actionable" if it was submitted and last updated during the
+        aircraft's first flight.
 
         Args:
             flights: Optional memoized flights for this target's user. If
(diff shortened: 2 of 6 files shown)


server/auvsi_suas/models/mission_config_test.py
```diff
@@ -251,17 +251,28 @@ def test_evaluate_teams(self):
         self.assertAlmostEqual(0.5, teams[user0]['uas_telem_times']['max'])
         self.assertAlmostEqual(1. / 6, teams[user0]['uas_telem_times']['avg'])
 
-        self.assertEqual(
-            11, teams[user0]['targets']['manual']['matched_target_value'])
+        # Target PK 5 matches target PK 1 (match value 0.28), Target PK 6
```
pmtischler: No need for this comment and the following one.

server/auvsi_suas/models/target.py
```diff
@@ -61,7 +62,7 @@ class TargetType(Choices):
     old data sets. Only add new values to the end.
     """
     standard = 1
-    qrc = 2
+    qrc = 2  # obsolete
```
pmtischler: Instead, file an issue for us to remove it and reference this line.

server/auvsi_suas/models/target.py
```diff
@@ -219,7 +220,7 @@ def json(self, is_superuser=False):
             'alphanumeric': alphanumeric,
             'alphanumeric_color': alphanumeric_color,
             'description': description,
-            'autonomous': self.autonomous,
+            'autonomous': self.autonomous
```
pmtischler: I'd keep the trailing comma

server/auvsi_suas/models/target.py
```diff
@@ -370,35 +372,32 @@ def match_value(self, submitted, real):
...
-            actionable_value = \
-                settings.TARGET_ACTIONABLE_PARAMS['objective']['value']
-        elif actionable == 'threshold':
-            actionable_value = \
-                settings.TARGET_ACTIONABLE_PARAMS['threshold']['value']
-        else:
-            actionable_value = 0
-
-        return classify_value + location_value + actionable_value
+            score = (float(settings.TARGET_LOCATION_THRESHOLD - location_dist)
```
pmtischler: Nit: I'd just combine this and location_value into one variable.

server/auvsi_suas/models/target.py
```diff
@@ -370,35 +372,32 @@ def match_value(self, submitted, real):
...
-                                           num_similar,
-                                           end_operator=operator.lt)
-        if not classify_value:
-            # Targets which don't have threshold classification have no value.
-            return 0
+        classify_value = (settings.CHARACTERISTICS_WEIGHT *
+                          real.similar_classifications(submitted))
 
         # Compute the location value.
         location_dist = float('inf')
```
pmtischler: I don't think you need this out of the if-statement scope anymore

server/auvsi_suas/models/target.py
```diff
@@ -274,13 +274,37 @@ def actionable_submission(self, flights=None):
...
             if flight.within(self.creation_time) and \
                 flight.within(self.last_modified_time):
                 return True
 
         return False
 
+    def interop_submission(self, missions=None):
+        """Checks if Target meets Interoperability submission criteria.
+
+        A target counts as being submitted over interoperability system if it
```
pmtischler: Nice.

server/auvsi_suas/models/target.py
```diff
@@ -438,27 +437,32 @@ def evaluation_dict(self):
...
+                unmatched_targets -= 1
 
                 classifications = submitted.similar_classifications(real)
                 location_accuracy = None
                 if submitted.location:
                     location_accuracy = submitted.location.distance_to(
                         real.location)
-                actionable = self.actionable(submitted, classifications,
-                                             location_accuracy)
+                actionable = float(submitted.actionable_submission(
```
pmtischler: These should probably be booleans, similar to `submitted.thumbnail_approved`.

pmtischler:
Agreed that the 2 missing changes are best in the scoring system as it does not impact target matching and are trivial to do in a spreadsheet or otherwise.

You should note that we need to provide an admin flag to the `interop_cli` to mark whether the upload is actionable, and accept such in the view code for admins only. We need this to upload USB submissions that were actionable.

LGTM. Make sure to squash the commits into 1 commit before submitting.

server/auvsi_suas/models/mission_config_test.py
yogishadixit: Done

server/auvsi_suas/models/target.py
yogishadixit: Done

server/auvsi_suas/models/target.py
yogishadixit: Done

server/auvsi_suas/models/target.py
yogishadixit: Done

server/auvsi_suas/models/target.py
yogishadixit: Done

server/auvsi_suas/models/target.py
yogishadixit: Done

pmtischler:
For #175

Issue #205, opened 2016-11-02
Remove QRC target type.


yogishadixit:
It is defined [here](https://github.com/auvsi-suas/interop/blob/master/server/auvsi_suas/models/target.py#L64).

PR #206, opened 2016-11-02
Removed QRC as target type.

pmtischler:
For #205


client/interop/types_test.py
```diff
@@ -365,13 +360,13 @@ def test_invalid(self):
         """Test invalid inputs."""
         # Bad latitude.
         with self.assertRaises(ValueError):
-            Target(type='qrc',
+            Target(type='emergent',
                    latitude='a',
                    longitude=-10,
                    description='http://test.com')
```
prattmic: Nit: change this to make sense for emergent target, just to be clear that this isn't ValueError because description contains a URL.

pmtischler: Done.

PR #207, opened 2016-11-03
Remove module inlining. Granular copy before setup.

pmtischler:
Removing module inlining removes the need to have all code files present
in order to import the module.

More granular copies means it doesn't redo the setup effort, like migrating
the database, every time any code changes.

The combination means making and testing code changes is much faster.
Waiting for a full database setup to test any small change was annoying.


prattmic:
LGTM

Issue #174, opened 2016-09-16
Waypoint capture & accuracy


pmtischler:
For best sequence, report accuracy of waypoints as `max(0, (100ft -distance) / 100ft)`.

Remove current waypoint grading (track-distance and capture threshold).

pmtischler:
Can this be closed?

PR #208, opened 2016-11-04
Update waypoint scoring

yogishadixit:
* Score waypoints using formula `max(0, (100ft -distance) / 100ft)`
* Report scores for best sequence
* Report closest approach to waypoints
* Update tests

### DIFF
--- a/server/auvsi_suas/models/mission_config.py
+++ b/server/auvsi_suas/models/mission_config.py
@@ -106,10 +106,10 @@ class MissionConfig(models.Model):
         Args:
             uas_telemetry_logs: A list of UAS Telemetry logs.
         Returns:
-            satisifed: Total number of waypoints hit (must be in order).
-            satisfied_track: Total number of waypoints hit while remaining
-                along the track. Once the UAV leaves the track no
-                more waypoints will count towards this metric.
+            best: Dictionary of distance to waypoints of the highest scoring
+                attempt.
+            scores: Dictionary of individual waypoint scores of the highest
+                scoring attempt.
             closest: Dictionary of closest approach distances to
                 each waypoint.
         """
@@ -121,33 +121,44 @@ class MissionConfig(models.Model):
 
         waypoints = self.mission_waypoints.order_by('order')
 
-        best = {"satisfied": 0, "satisfied_track": 0}
-        current = {"satisfied": 0, "satisfied_track": 0}
+        best = {}
+        current = {}
         closest = {}
+
+        def score_waypoint(distance):
+            """Scores a single waypoint."""
+            return max(
+                0, float(settings.SATISFIED_WAYPOINT_DIST_MAX_FT - distance) /
+                settings.SATISFIED_WAYPOINT_DIST_MAX_FT)
+
+        def score_waypoint_sequence(sequence):
+            """Returns scores given distances to a sequence of waypoints."""
+            score = {}
+            for i in range(0, len(waypoints)):
+                score[i] = \
+                    score_waypoint(sequence[i]) if i in sequence else 0.0
+            return score
 
         def best_run(prev_best, current):
             """Returns the best of the current run and the previous best."""
-            if current["satisfied"] > prev_best["satisfied"]:
-                return current
-            elif current["satisfied"] == prev_best["satisfied"] and \
-                    current["satisfied_track"] > prev_best["satisfied_track"]:
-                return current
-            return prev_best
+            prev_best_scores = score_waypoint_sequence(prev_best)
+            current_scores = score_waypoint_sequence(current)
+            if sum(current_scores.values()) > sum(prev_best_scores.values()):
+                return current, current_scores
+            return prev_best, prev_best_scores
 
         prev_wpt, curr_wpt = -1, 0
-        track_ok = True
 
         for uas_log in uas_telemetry_logs:
             # At any point the UAV may restart the waypoint pattern, at which
             # point we reset the counters.
             d0 = uas_log.uas_position.distance_to(waypoints[0].position)
             if d0 < settings.SATISFIED_WAYPOINT_DIST_MAX_FT:
-                best = best_run(best, current)
+                best = best_run(best, current)[0]
 
-                # The check below will set satisfied and satisfied_track to 1.
-                current = {"satisfied": 0, "satisfied_track": 0}
+                # Reset current to default values.
+                current = {}
                 prev_wpt, curr_wpt = -1, 0
-                track_ok = True
 
             # The UAS may pass closer to the waypoint after achieving the capture
             # threshold. so continue to look for better passes of the previous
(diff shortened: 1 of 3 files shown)


server/auvsi_suas/models/mission_config_test.py
```diff
@@ -107,38 +107,33 @@ def test_satisfied_waypoints(self):
...
         def assertSatisfiedWaypoints(expect, got):
             """Assert two satisfied_waypoints return values are equal."""
-            self.assertEqual(expect[0], got[0])
-            self.assertEqual(expect[1], got[1])
-            for k in expect[2].keys():
-                self.assertIn(k, got[2])
-                self.assertAlmostEqual(expect[2][k], got[2][k], delta=0.1)
-            for k in got[2].keys():
-                self.assertIn(k, expect[2])
+            assertDictionaryEqual(expect[0], got[0])
```
pmtischler: Why did you add a custom assert? Is it to have a threshold for floats?

server/auvsi_suas/models/mission_config.py
```diff
@@ -121,16 +121,18 @@ def satisfied_waypoints(self, uas_telemetry_logs):
 
         waypoints = self.mission_waypoints.order_by('order')
 
-        best = {"satisfied": 0, "satisfied_track": 0}
-        current = {"satisfied": 0, "satisfied_track": 0}
+        best = {i: 0.0 for i in range(0, len(waypoints))}
```
pmtischler: This function had a bad assumption that we need to fix now that track distance has been removed. The assumption was that quads would never go backwards through waypoints (likely true), and that planes might want to but couldn't without violating track distance (likely true). Consider what happens now if a plane decides to attempt flying past a waypoint multiple times to attempt a higher score, but happens to touch the next waypiont while turning around. In the past, such would violate the track boundary for sure, but now it'd be valid but mis-scored. The greedy approach here probably won't be able to account for such. You probably need to do a dynamic programming approach using all points within waypoint threshold. That being said, I'd prefer to submit this first and send a follow-up PR with a test case for this edge case and the new implementation.

For each time-sorted point, the dynamic algorithm could be:

for each waypoint:
  for each data point:
    if first waypoint:
      set value of v[waypoint, datapoint] to score_waypoint()
      set value of sv[waypoint, datapoint] to the same
      continue
    get previous waypoint value as max(prev waypoint, prev datapoint)
    set value of v[waypoint, datapoint] to 
          max(score_waypoint(cur waypoint, cur datapoint) + v[prev waypoint, prev datapoint],
                  v[cur waypoint, prev datapoint])
return max over all of d, trace back to get closest on best path

For N waypoints and M data points, where M >> N, runtime would be O(NM).

You can further optimize by only tracking/computing/iterating over data points relevant for a waypoint (ones under threshold).

pmtischler: Updated the prev comment removing some cruft from the psuedocode.

server/auvsi_suas/models/mission_config.py
```diff
@@ -271,11 +253,10 @@ def evaluate_teams(self, users=None):
...
 
             # Determine if the uas hit the waypoints.
-            waypoints_hit, waypoints_hit_track, waypoints_closest = \
+            waypoint_scores, closest_approaches = \
                 self.satisfied_waypoints(uas_logs)
-            eval_data['waypoints_satisfied'] = waypoints_hit
-            eval_data['waypoints_satisfied_track'] = waypoints_hit_track
-            eval_data['waypoints_closest'] = waypoints_closest
+            eval_data['waypoint_scores'] = waypoint_scores
+            eval_data['waypoint_closest_approaches'] = closest_approaches
```
pmtischler: That is, we want to report they closest they got over all time, and the closest they got on the highest scoring sequence. To prevent cross-system scoring inconsistency, we want to keep waypoint_scores.

pmtischler:
Don't forget to squash before submit. Also, use "For #x" in the description so it gets referenced in the issue. When the PR fixes it, you can use Fixes instead of For and it will auto close the issue for you. This PR doesn't fix yet (see comments).

Let's also have @prattmic review.
 
This PR LGTM for merging. Left note about a follow-up PR to handle a new-ish edge case.

server/auvsi_suas/models/mission_config_test.py
yogishadixit: Yes. The provided assertDictEquals doesn't do float comparison well.

server/auvsi_suas/models/mission_config_test.py
```diff
@@ -107,38 +107,33 @@ def test_satisfied_waypoints(self):
         user = User.objects.create_user('testuser', 'testemail@x.com',
                                         'testpass')
 
+        def assertDictionaryEqual(expect, got):
```
prattmic: `def assertDictionaryEqual(self, expect, got):`

I'm not sure why assertSatisifiedWaypoints doesn't pass self either.

server/auvsi_suas/models/mission_config_test.py
```diff
@@ -107,38 +107,33 @@ def test_satisfied_waypoints(self):
...
             """Assert two satisfied_waypoints return values are equal."""
-            self.assertEqual(expect[0], got[0])
-            self.assertEqual(expect[1], got[1])
-            for k in expect[2].keys():
-                self.assertIn(k, got[2])
-                self.assertAlmostEqual(expect[2][k], got[2][k], delta=0.1)
-            for k in got[2].keys():
-                self.assertIn(k, expect[2])
+            assertDictionaryEqual(expect[0], got[0])
+            assertDictionaryEqual(expect[1], got[1])
```
prattmic: `self.assertDictionaryEqual(expect[1], got[1])`

server/auvsi_suas/models/mission_config.py
```diff
@@ -121,16 +121,18 @@ def satisfied_waypoints(self, uas_telemetry_logs):
```
prattmic: The docstring above needs to be updated.

(Why doesn't Github let me leave comments anywhere in the file??)

yogishadixit:
Added `waypoints_closest_for_scores`, which changed code structure a bit. PTAL, thanks!

server/auvsi_suas/models/mission_config.py
```diff
@@ -106,10 +106,10 @@ def satisfied_waypoints(self, uas_telemetry_logs):
         Args:
             uas_telemetry_logs: A list of UAS Telemetry logs.
         Returns:
-            satisifed: Total number of waypoints hit (must be in order).
-            satisfied_track: Total number of waypoints hit while remaining
-                along the track. Once the UAV leaves the track no
-                more waypoints will count towards this metric.
+            best: Dictionary of distances to waypoints of the highest scoring
```
pmtischler: distance

yogishadixit:
Fixes #174.

PR #212, opened 2016-12-07
Add a team_id field to targets POST

yogishadixit:
It will allow an admin user to create a target on behalf of a team. This change is a prereq to adding an `actionable_override` flag.

### DIFF
--- a/client/interop/types.py
+++ b/client/interop/types.py
@@ -45,7 +45,7 @@ class ClientBaseType(object):
                 serial[attr] = data.serialize()
             elif isinstance(data, list):
                 serial[attr] = [d.serialize() for d in data]
-            else:
+            elif data is not None:
                 serial[attr] = data
         return serial
 
@@ -252,6 +252,8 @@ class Target(ClientBaseType):
             certain target types.
         autonomous: Optional; defaults to False. Indicates that this is an
             ADLC target.
+        team_id: Optional. The username of the team on whose behalf to submit
+            targets. Must be admin user to specify.
 
     Raises:
         ValueError: Argument not valid.
@@ -259,7 +261,7 @@ class Target(ClientBaseType):
 
     attrs = ['id', 'user', 'type', 'latitude', 'longitude', 'orientation',
              'shape', 'background_color', 'alphanumeric', 'alphanumeric_color',
-             'description', 'autonomous']
+             'description', 'autonomous', 'team_id']
 
     def __init__(self,
                  id=None,
@@ -273,7 +275,8 @@ class Target(ClientBaseType):
                  alphanumeric=None,
                  alphanumeric_color=None,
                  description=None,
-                 autonomous=False):
+                 autonomous=False,
+                 team_id=None):
         self.id = id
         self.user = user
         self.type = type
@@ -286,3 +289,4 @@ class Target(ClientBaseType):
         self.alphanumeric_color = alphanumeric_color
         self.description = description
         self.autonomous = autonomous
+        self.team_id = team_id
--- a/client/tools/interop_cli.py
+++ b/client/tools/interop_cli.py
@@ -30,7 +30,7 @@ def targets(args, client):
             raise ValueError('--target_dir is required.')
         upload_legacy_targets(client, args.legacy_filepath, args.target_dir)
     elif args.target_dir:
-        upload_targets(client, args.target_dir)
+        upload_targets(client, args.target_dir, args.team_id)
     else:
         targets = client.get_targets()
         for target in targets:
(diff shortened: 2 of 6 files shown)


client/interop/client.py
```diff
@@ -198,7 +201,9 @@ def post_target(self, target):
             requests.Timeout: Request timeout.
             ValueError or AttributeError: Malformed response from server.
         """
-        r = self.post('/api/targets', data=json.dumps(target.serialize()))
+        serialized_target = target.serialize()
```
pmtischler: I would not modify this function. It'd be better to have Target object take a team_id field.

client/tools/interop_cli.py
```diff
@@ -123,7 +126,7 @@ def main():
         password = getpass.getpass('Interoperability Password: ')
 
     # Create client and dispatch subcommand.
-    client = Client(args.url, args.username, password)
+    client = Client(args.url, args.username, password, args.team_id)
```
pmtischler: I wouldn't modify the client API for a data field for a single request.

client/tools/interop_cli.py
```diff
@@ -73,6 +73,9 @@ def main():
                         required=True,
                         help='Username for interoperability.')
     parser.add_argument('--password', help='Password for interoperability.')
+    parser.add_argument('--team_id',
```
pmtischler: This is specific to targets, lets add it to the target subparser.

server/auvsi_suas/views/targets.py
```diff
@@ -145,6 +145,14 @@ def post(self, request):
         if 'type' not in data:
             return HttpResponseBadRequest('Target type required.')
 
+        # Team id required if submitting target as superuser.
+        if 'team_id' not in data:
```
pmtischler: I would not require team_id. This should be an optional override that can only be specified by admins. Otherwise each team will need to specify their team ID with each request, which they've already done by logging in and passing the cookie. I'd prefer to not change the public API teams interact with. I'd like this to be an undocumented admin-only feature that only we use for Object File Format upload.

yogishadixit: What should the behavior be if a team passes in team_id anyway. Should I just ignore it, or return bad request?

server/auvsi_suas/views/targets.py
```diff
@@ -149,7 +149,8 @@ def post(self, request):
         user = request.user
         if request.user.is_superuser:
             if 'team_id' not in data:
-                return HttpResponseBadRequest('Team ID required for superuser.')
+                return HttpResponseBadRequest(
```
pmtischler: I think you want an error to be expressed as
```
if 'team_id' in data and not request.user.is_superuser:
```

That is, admins can specify an optional override- not that they have to.

server/auvsi_suas/views/targets.py
pmtischler: Return 403- Forbidden (not authorized)

server/auvsi_suas/views/targets.py
yogishadixit: Why shouldn't it be the case that admins must provide `team_id`? Can/do admins create targets?

pmtischler: Yes, like for the mission's correct targets used for grading.

pmtischler:
Travis failure for client test.

client/interop/types.py
```diff
@@ -252,14 +252,16 @@ class Target(ClientBaseType):
             certain target types.
         autonomous: Optional; defaults to False. Indicates that this is an
             ADLC target.
+        team_id: Optional. The username of the team on whose behalf to submit
+            targets.
```
pmtischler: Also add: Must be admin user to specify.

client/tools/upload_targets.py
```diff
@@ -132,18 +132,21 @@ def upload_legacy_targets(client, target_filepath, imagery_dir):
...
-def upload_target(client, target_file, image_file):
+def upload_target(client, target_file, image_file, team_id=None):
     """Upload a single target to the server
 
     Args:
         client: interop.Client connected to the server
         target_file: Path to file containing target details in the Object
             File Format.
         image_file: Path to target thumbnail. May be None.
+        team_id: The username of the team on whose behalf to submit targets.
```
pmtischler: Here too :)

server/auvsi_suas/views/targets.py
```diff
@@ -145,6 +145,15 @@ def post(self, request):
         if 'type' not in data:
             return HttpResponseBadRequest('Target type required.')
 
+        # Team id required if submitting target as superuser.
```
pmtischler: Need to fix comment, e.g.
Team ID can only be specified if superuser.

pmtischler:
LGTM

pmtischler:
Ping. What is the status?

PR #213, opened 2016-12-07
Add urllib3+requests to tools requirements for docs build.

pmtischler:
Appears that urllib3+requests has become a new dependency as builds have been
failing with:
Could not import extension sphinx.builders.linkcheck (exception: No
module named packages.urllib3.exceptions)


yogishadixit:
LGTM

PR #214, opened 2016-12-13
Interop CLI actionable override flag

yogishadixit:


server/auvsi_suas/models/target.py
```diff
@@ -271,14 +276,14 @@ def actionable_submission(self, flights=None):
...
+        actionable = False
         if len(flights) > 0:
             flight = flights[0]
             if flight.within(self.creation_time) and \
                 flight.within(self.last_modified_time):
-                return True
+                actionable = True
 
-        return False
+        return actionable or self.actionable_override
```
pmtischler: I would just check self.actionable_override first

pmtischler:
LGTM

pmtischler:
Note travis failure regarding style of migration.

pmtischler:
Ping. What is the status?

server/auvsi_suas/models/target.py
```diff
@@ -259,8 +261,11 @@ def similar_classifications(self, other):
     def actionable_submission(self, flights=None):
         """Checks if Target meets Actionable Intelligence submission criteria.
 
-        A target is "actionable" if it was submitted and last updated during the
-        aircraft's first flight.
+        A target is "actionable" if one of the following conditions is met:
+            (a) If it was submitted over interop and last updated during the
+                aircraft's first flight.
+            (b) If the target was submitted via USB, the target's
+                actionable_override flag is set to True.
```
pmtischler: Nit: If the target's actionable_override was set by an admin.

server/auvsi_suas/views/targets.py
```diff
@@ -115,6 +115,11 @@ def normalize_data(data):
         if data['autonomous'] is not True and data['autonomous'] is not False:
             raise ValueError('"autonmous" must be true or false')
 
+    if 'actionable_override' in data:
+        if (data['actionable_override'] is not True and
+                data['actionable_override'] is not False):
```
pmtischler: Nit: can you align the "data"

pmtischler: Does this check work with non True/False values, but which have a truth value?

pmtischler:
LGTM after comments.

pmtischler:
To clarify, you still need to update the CLI, correct?

pmtischler:
Any update on this?

server/auvsi_suas/views/targets.py
yogishadixit: Works for 0, but not any of the others. Tbh, I just copied the logic on line 115.

yogishadixit: Also, in JSON, a boolean value can only be specified as True or False, so specifying any other truth values shouldn't be valid, right?

client/tools/upload_targets.py
```diff
@@ -166,6 +173,8 @@ def upload_targets(client, target_dir, team_id=None):
             File Format and target thumbnails.
         team_id: The username of the team on whose behalf to submit targets.
             Defaults to None.
+        actionable_override: Manually sets the target to be actionable. Defaults
```
pmtischler: Optional. Overrides the target as actionable. Must be superuser to set.

server/auvsi_suas/views/targets.py
```diff
@@ -163,6 +168,11 @@ def post(self, request):
             return HttpResponseBadRequest(
                 'Either none or both of latitude and longitude required.')
 
+        # Cannot submit JSON with actionable_override if not superuser.
+        if not request.user.is_superuser and 'actionable_override' in data:
```
pmtischler: Nit: Maybe swap the order of if statements, I think it reads better.
```
if 'actionable_override' in data and not request.user.is_superuser:
```

server/auvsi_suas/views/targets.py
```diff
@@ -245,6 +256,11 @@ def put(self, request, pk):
         except ValueError:
             return HttpResponseBadRequest('Request body is not valid JSON.')
 
+        # Cannot submit JSON with actionable_override if not superuser.
+        if not request.user.is_superuser and 'actionable_override' in data:
```
pmtischler: Same here.

pmtischler:
LG, just some minor nits. Feel free to submit after any updates (once travis passes).

client/tools/interop_cli.py
```diff
@@ -111,7 +112,11 @@ def main():
     subparser.add_argument(
         '--team_id',
         help='''The username of the team on whose behalf to submit targets.
-Must be admin user to specify''')
+Must be admin user to specify.''')
+    subparser.add_argument(
+        '--actionable_override',
```
pmtischler: type=bool, default=False

pmtischler: Oh I see, it uses None to not set, SGTM. Setting the flag passes a `true` value?

yogishadixit: Yep, according to docs here: https://docs.python.org/2/howto/argparse.html#introducing-optional-arguments

PR #219, opened 2017-01-30
Set default logging to INFO. Sleep before client test.

pmtischler:
For high-rate telemetry forwarding, we don't want to print urllib debug messages.

Sleeping will reduce flakiness of client test and spurious build errors.


pmtischler:
This PR is trivial and it's been a few days. I'm going to submit. I can send a follow-up PR with changes if need be.

PR #223, opened 2017-02-28
Add healthcheck to container, poll in Travis.

pmtischler:
Purpose is to decrease sporatic CI test failures because the client
tests start before the server is ready. Secondary benefit to increase
health status visibility via a secondary prober.

Healthcheck which will probe the Docker container to validate that the
server is healthy. It simply polls the homepage (/) using curl. It
stores the status and any errors, and can be queried using `docker
inspect`. Format is defined here:
https://docs.docker.com/engine/reference/builder/#/healthcheck

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,10 @@ install:
   - ./client/build.sh
 
 script:
-  # Test server & client.
-  - ./server/run.sh
+  # Test server.
   - ./server/test.sh
-  - sleep 10  # Sleep to allow server to start.
+  - ./server/run.sh
+  # Test client.
   - ./client/test.sh
   # Verify that docs build.
   - ./tools/build_docs.sh
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -5,6 +5,7 @@ WORKDIR /interop/server
 RUN sudo apt-get -qq update && sudo apt-get -qq install -y \
         apache2 \
         apache2-utils \
+        curl \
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -13,10 +13,12 @@ install:
   - ./client/build.sh
 
 script:
-  # Test server & client.
+  # Test server.
   - ./server/run.sh
   - ./server/test.sh
-  - sleep 10  # Sleep to allow server to start.
+  # Poll server for healthiness before proceeding.
+  - for i in {1..10}; do docker inspect interop-server | grep healthy && break || sleep 10; done
```
prattmic: Consider putting this in run.sh.

Also, fail the test if it never comes up?

prattmic: It would be cleaner to extract exactly what you want. I think `docker inspect -f "{{.State.Health.Status}}"` is the right field.

pmtischler: Done.

pmtischler: Done.

prattmic:
`grep healthy` will also match 'unhealthy'

pmtischler:
Haha oops. I'll send a fix in later today.

pmtischler:
Fixed in [5e8f808](https://github.com/auvsi-suas/interop/commit/5e8f808f44f0ecf471cee7c7f01cb5a75a754dc7)

Issue #218, opened 2017-01-29
Automatic scoring of mission


pmtischler:
* Store object keyed by (mission, team) storing manual judge data
* Calculating final mission score from interop and manual data

PR #225, opened 2017-03-07
Return csv/json team evaluations as single zip.

pmtischler:
A single zip file makes it more efficient to get both json and csv back
at the same time (csv for higher-level scoring, json for team feedback).
Per team json makes creating a single PDF for feedback easier as it
doesn't require parsing the json to break out by team.

For #218


pmtischler:
Ping.

pmtischler:
Short of comments, I'm going to submit tomorrow morning so I can continue working on that issue.

PR #230, opened 2017-03-16
Use proto to serialize. Improve auto-target.

pmtischler:
Use protocol buffers to create data objects and serialize to json as
part of team evaluation. This saves code, adds minor type checking, and
is likely more robust than our custom methods. Post-competition, we may
want to do similar to our team-facing API and client libraries.

To handle manual+auto submissions for the same target, the previous
implementation took the max score across, and counted extra submissions
for each separatley. This assumes you wouldn't match differently if you
considered them simultaneously, which may not be correct. Doing it
simultaneously complicates extra-object counting, as now you need to
detect when manual/auto could have replaced each other. This adds an
extra bipartite-matching for counting what could have been valid unused
matches, so they are counted to unmatched.


pmtischler:
For #218. Filed #231

pmtischler:
For the same of making progress, I'm going to submit Monday morning in lack of comments.

PR #235, opened 2017-03-25
Add judge feedback to evaluation.

pmtischler:
For #218


pmtischler:
Sans-comments, I plan to submit on Monday.

PR #249, opened 2017-04-18
Waypoint interpolation

yogishadixit:
Implement linear + spline interpolation for waypoints. Spline interpolation takes a really long time to run, so I would appreciate any advice.

### DIFF
--- a/server/auvsi_suas/models/uas_telemetry.py
+++ b/server/auvsi_suas/models/uas_telemetry.py
@@ -7,6 +7,7 @@ from django.utils import timezone
 
 from access_log import AccessLog
 from aerial_position import AerialPosition
+from auvsi_suas.models.gps_position import GpsPosition
 from auvsi_suas.models import distance
 from auvsi_suas.models import units
 from auvsi_suas.patches.simplekml_patch import AltitudeMode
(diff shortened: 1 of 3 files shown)


server/auvsi_suas/models/uas_telemetry.py
```diff
@@ -255,12 +259,59 @@ def best_run(prev_best, current):
                 return current, current_scores
             return prev_best, prev_best_scores
 
+        def spline_interpolated_distance(telems, wpt):
```
pmtischler: This function isn't used. Also, as we talked it's probably best to stick to linear interpolation. Please drop this and update the imports.

server/auvsi_suas/models/uas_telemetry.py
```diff
@@ -255,12 +259,59 @@ def best_run(prev_best, current):
...
+                return apos.distance_to(wpt.position)
+            # TODO(Yogisha): Figure out how to make this more efficient.
+            # Currently takes >10 minutes to finish.
+            min_ts = fmin_cobyla(
+                objective,
+                x0=[calendar.timegm(telem[0].timestamp.timetuple())],
+                cons=[])
+            return objective(min_ts)
+
+        def linearly_interpolated_distance(telem_start, telem_end, wpt):
```
pmtischler: Please add unit tests to verify correctness of interpolation, and that it doens't interpolate when over speed.

server/auvsi_suas/models/uas_telemetry.py
```diff
@@ -255,12 +259,59 @@ def best_run(prev_best, current):
...
+            def uas_position_to_tuple(pos):
+                return (pos.gps_position.latitude,
+                        pos.gps_position.longitude,
+                        pos.altitude_msl)
+
+            # Check if data is too weird to use interpolation.
+            d = telem_start.uas_position.distance_to(telem_end.uas_position)
+            t = (telem_end.timestamp - telem_start.timestamp).total_seconds()
+            if (d / t) > 36: # max allowed speed in m/s
```
pmtischler: Please move 36 to settings file. Also, distance is calculated in feet, so d/t is feet per second.

pmtischler:
The files aren't formatted correctly, please run the tool

server/auvsi_suas/models/uas_telemetry.py
pmtischler: Please also add a small unit test to verify interpolation at the satisfied_waypoints level.

server/auvsi_suas/models/uas_telemetry.py
pmtischler: To clarify, the reason to avoid splines is how it deals with noise or points which are too close to each other. There can be weird effects, and at 1 Hz the difference shouldn't matter.

server/auvsi_suas/models/uas_telemetry.py
pmtischler: Also, you need to check that t < 1.0 (they need to meet min rate qualifications before we give benefit of doubt). For example, you wouldn't want to interpolate points 10 minutes apart.

server/auvsi_suas/models/uas_telemetry.py
pmtischler: Please add a unit test to verify 1Hz handling as well.

pmtischler:
The commit seems to be rooted in the previous commit:
1) Let's use 1.5sec as a threshold for interpolation, and add a test, with value in settings file
2) You need to run the formatter (per the Travis failure)
3) Looks like your branch has conflicts, a rebase might be in order

server/auvsi_suas/models/uas_telemetry.py
```diff
@@ -206,6 +207,38 @@ def live_kml(cls, kml, timespan):
...
+            waypoint: The waypoint for which to find the closest distance.
+            utm: The UTM Proj projection to project into.
+        Returns:
+            The closest distance to the waypoint in feet.
+        """
+        # Verify that aircraft velocity is within bounds. Don't interpolate if
+        # it isn't because the data is probably erroneous.
+        d = start_log.uas_position.distance_to(end_log.uas_position)
+        t = (end_log.timestamp - start_log.timestamp).total_seconds()
+        if (t > settings.MAX_TELMETRY_INTERVAL_SEC or
```
pmtischler: Nit: Add interpolate word somewhere in constant name.

pmtischler:
LGTM

PR #254, opened 2017-05-05
Update waypoint scoring to 2017 rules.

### LINKED ISSUE #250: Waypoint Navigation Evaluation Bug
On a fresh install. I created a Mission Config object with 3 waypoints. I simulated a waypoint navigation by creating 3 UAS Telemetry objects using the Django Admin interface. The 3 UAS telemetry objects had the exact same latitude, longitude, and altitude as the mission waypoints. Then I evaluated. As shown in the screenshot below, full points were given for all 3 waypoints. This is the expected behavior.

waypoint_navigation_evaluation_hit_waypoints_123

However, I then deleted the 1st UAS Telemetry object and re-evaluated. As shown in the screenshot below, zero points were given for all 3 waypoints. This is not the expected behavior, because only the 1st waypoint was missed. The 2nd and 3rd waypoints were hit perfectly. Full points should have been given for the 2nd and 3rd waypoints.

waypoint_navigation_evaluation_hit_waypoints_23

At least I believe that is not the expected behavior, based on the rulebook. In the Mission Demonstration, Autonomous Flight, Waypoint Accuracy section, it says "Each waypoint will be weighted equally, and the ratio of points received per waypoint will be max(0, (100ft − distance) / 100ft)".

Thanks.

pmtischler:
In previous rules, you had to travel within 100ft and it was assumed
that you couldn't skip between waypoints as a result. In the new rules,
the constraint is only that they waypoint hits used must be in sequence,
and the max scoring sequence is used.

This new approach uses dynamic programming to find the best sequence. It
first filters the telemetry logs for the hits (much fewer) and then
calculates the max-scoring sequence recurrence.

Fixes #250

### DIFF
--- a/server/auvsi_suas/models/uas_telemetry.py
+++ b/server/auvsi_suas/models/uas_telemetry.py
@@ -1,21 +1,23 @@
 """UAS Telemetry model."""
 
+import itertools
+from collections import defaultdict
+
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.db import models
 from django.utils import timezone
 
-from access_log import AccessLog
-from aerial_position import AerialPosition
-from auvsi_suas.models.gps_position import GpsPosition
 from auvsi_suas.models import distance
 from auvsi_suas.models import units
+from auvsi_suas.models.access_log import AccessLog
+from auvsi_suas.models.aerial_position import AerialPosition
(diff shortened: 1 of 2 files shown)


pmtischler:
Sans comments, I will submit tomorrow morning.

PR #278, opened 2017-07-12
Use filter_vertical for ManyToManyFields in admin

### LINKED ISSUE #272: Consider using filter_horizontal or filter_vertical for ManyToManyFields in admin
Would it be possible to use [`filter_horizontal`](https://docs.djangoproject.com/en/dev/ref/contrib/admin/#django.contrib.admin.ModelAdmin.filter_horizontal) or [`filter_vertical`](https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#django.contrib.admin.ModelAdmin.filter_vertical) for the `ManyToManyFields` in the admin?

That would make it easier to select mission waypoints, select search grid points and select targets, and would also make it a lot easier for new users to distinguish between the available options and the selected options.

legonigel:
Also add some custom CSS to the admin page to allow the filter
vertical to scroll, becasue by default on chrome they do not scroll horizontally and
many of the models have very long string representations.

Closes #272

### DIFF
--- a/server/auvsi_suas/admin.py
+++ b/server/auvsi_suas/admin.py
@@ -17,10 +17,19 @@ from auvsi_suas.models.waypoint import Waypoint
 class LargeDataModelAdmin(admin.ModelAdmin):
     show_full_result_count = False
 
+
+# Define model admin which has scrollable select boxes
+# By default on chrome select has overflow-x: hidden
+class ScollableFilterModelAdmin(admin.ModelAdmin):
+    class Media:
+        css = {"all": ("auvsi_suas/admin/scroll_filter.css", )}
+
 # Register models for admin page
 
 # Only display raw ID fields for ForeignKeys which may contain large amounts
 # of data.
+
+# Use filter_horizontal for ManyToMany fields to easier choose items
 
 
 @admin.register(AerialPosition)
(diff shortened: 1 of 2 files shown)


legonigel:
Before:
image

After:
image

pmtischler:
Does it not scroll because it fills the body and the body can scroll?

legonigel:
It doesn't fill the body or scroll horizontally by default, which makes some data not visible at all. 
This might not apply to all browsers but by default [chromium applies `overflow-x:hidden` to selects](https://chromium.googlesource.com/chromium/blink/+/master/Source/core/css/html.css#777).
It could be fixed by manually specifying a width for the select as well, but I would have to investigate how that throws off other elements, because the select's div is floating.

legonigel:
This is what it looks like without the custom CSS on Chromium, on FireFox the layout looks identical.
image

pmtischler:
Sgtm

foochuanwei:
@legonigel 'filter_verical' -> 'filter_vertical'?

pmtischler:
Made unnecessary by #281

In the future, please don't send a PR if you're not going to go through a timely review process.

legonigel:
@pmtischler What else was I expected to do to that pull request? What caused this pull request to be closed instead of merged?

PR #347, opened 2018-10-04
Remove analysis not used in scoring.

### LINKED ISSUE #343: Remove all analysis/feedback not used in scoring or display
Reduce volume of code, processing time, and false signals by removing analysis/feedback not used.

pmtischler:
Simplifies the code base and removes confusion from conflicting interop
analysis and judge feedback.

Out of bounds is noisy due to handling of takeoff/landing regions, and
prevents us from having to create extra fly zones covering the area
around the takeoff zone which confused teams.

Mission time is perfectly accurate at the lead judge (e.g. sees the team
enter/leave runway), but interop operator usually gets delayed
information, usually leading to conflicting values.

Fixes #343

### DIFF
--- a/server/auvsi_suas/models/mission_evaluation.py
+++ b/server/auvsi_suas/models/mission_evaluation.py
@@ -78,26 +78,8 @@ def generate_feedback(mission_config, user, team_eval):
     """
     feedback = team_eval.feedback
 
-    # Calculate the total mission clock time.
-    missions = MissionClockEvent.missions(user)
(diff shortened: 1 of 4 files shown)


pmtischler:
Note that the same delay applies to takeoff/landing events, but it's not as bad (operator can see/hear when plane takesoff/lands), and doesn't need to be as accurate and just reasonably close (shouldn't impact rate calc, before reaching waypoint area, and before approaching obstacles).

PR #391, opened 2019-04-19
Fix missing shape color field on ODLC review.

mattr555:
Testing our integration with the new proto-based api, I was afraid we did something wrong when I checked the ODLC review screen and saw no shape color:

Screenshot_2019-04-19_01-50-23

Turns out the issue was actually with the interop server.


pmtischler:
LGTM, thanks!

PR #441, opened 2019-09-02
Remove OFF_AXIS as a type of ODLC.

### LINKED ISSUE #437: Drop OFF_AXIS as a type for ODLC in API
`OFF_AXIS` is really a `STANDARD` where the position is provided already. Unnecessarily complicates scoring.

pmtischler:
OFF_AXIS is the same as STANDARD, except we don't give points if it's
not differentiated correctly. Removing the type specialization will be
more flexible, and not double penalize a detection not having GPS that
can be labeled out of flight boundary.

Fixes #437


pmtischler:
Note to teams which may later read this:

There is still an off-axis ODLC at competition, the position of which is given out by interop as part of downloading mission details. However, when uploading the details of that ODLC to interop, you now specify the type as STANDARD. This will now give a team points for ODLC if they detect it but failed to mark it as the OFF_AXIS.

PR #458, opened 2019-10-05
Pull images so they're available in cache-from.

pmtischler:


pmtischler:
Did the trick- CI went from 12min21sec to 3min30sec.

PR #476, opened 2020-03-03
Mission creation automation scripts

Perseus101:
I saw a question on the google group a few days ago about setting up automation scripts for mission creation, so I thought I'd offer up one of the simple convenience scripts we use internally. It's a convenient script, and provides an easy starting point for creating further automation scripts.

The `create_mission.py` script takes a JSON file in the same format as the output from the `/mission/{id}` endpoint and creates a new mission, prompting for any missing required fields. I've included a full example mission in `sample_mission,json` for easy setup. The sample mission is identical to the sample data from the `test_utils` mission.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -1,0 +1,12 @@
+# Running mission creation scripts
+```
+$ docker-compose up
+... new terminal ...
+$ docker-compose run interop-server python3 /interop/scripts/create_mission.py -m /interop/scripts/sample_mission.json
+```
+
+# Interacting with the server via curl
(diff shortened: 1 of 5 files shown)


bretonjustin:
Thanks man!

pmtischler:
Sorry, but we don't accept contrib stuff as we don't have the cycles to maintain it. Thank you for helping other teams- I'd recommend uploading this to a team-owned GitHub repo and then replying on that Groups thread with the link.

PR #488, opened 2020-06-05
Switch hardcoded packages to find_packages()

coderkalyan:
I found some issues with the subpackages (client, proto) not being properly installed with `python setup.py install`. Switching the hardcoded `packages=['auvsi-suas']`to setuptools.find_packages() fixes this, and nevertheless is good practice anyway!


pmtischler:
Thanks!

PR #489, opened 2020-06-10
Use /usr/bin/env in all shebangs

elihunter173:
`interop-server.sh` doesn't run on NixOS (and I think probably not Mac but I can't confirm) because of the shebang being `#!/bin/bash` instead of `#!/usr/bin/env bash`. This switches all scripts' shebangs to use `/usr/bin/env`.


pmtischler:
Thanks!

PR #525, opened 2021-12-30
Fix calculation of interop telem rates.

### LINKED ISSUE #465: Scoring and Telemetry problem
At first I installed last year's competition interop server and tested everything was cool.Then I removed all docker images and containers  installed latest interop. At the latest  interop I get  autonomous flight, obstacle avoidance, waypoint accuracy points as 0 and telemetryprequisite false .How can I fix this problem ?  
Also when I change an existing takeofforlandingevent and then evaluate I get "Infinite flight period, may be missing TakeoffOrLandingEvent." warning.When I add new events 1 for takeoff 1 for landing I get max and telemetry seconds as:
 "uasTelemetryTimeMaxSec": 314.0,
 "uasTelemetryTimeAvgSec": 314.0,

pmtischler:
Broken in #439 which introduced a double use of generator.

Fixes #465.


manavgandhi17:
I can confirm that this commit has fixed the issue on my end as well.

PR #527, opened 2021-12-31
Repeatedly print telemetry rate.

### LINKED ISSUE #526: Readout of Current Telemetry Rate (mavlink_proxy)
Good Evening,

After splitting my mavlink stream via mavproxy and using the included client-side package for sending information to the server - I have gained successful telemetry communication. I can actively see the telemetry being received on the server-side and this is confirmed during team evaluation where the average telemetry rate is roughly 0.25 seconds. Thus, I can confirm that this portion is working.

However, unlike past times where I have used this setup (it has been over a year since I have run through this process) - the mavlink telemetry rate does not continue to printout after the first time.

image

This isn't critical to making sure the whole system works but was a nice way of knowing that the telemetry was being sent at a reasonable rate during competition (w/o access to client-side). Any help in figuring out this issue would be greatly appreciated.

pmtischler:
Fixes #526


manavgandhi17:
This has fixed the issue. Thanks!

For what it's worth the .README does not go over the "step" for updating the client-side Docker image. It is a simple one-liner based on the provided bash script (just changing run to pull), but would probably be nice to have explicitly stated in the .README for newer users.

